PR #267, opened 2015-02-17
WinForms: Makes ListBox support TextBinding and ImageBinding

bchavez:
Compiles and visually tested with `Eto.Test.WinForms.exe`


cwensley:
Awesome, very nicely done!!

One small thing - could you make it check that the bindings are not null before using them?  Other platforms just check for the ImageBinding being null, but they really should be checking both.

cwensley:
Beautiful!  Thanks for the fix!

PR #269, opened 2015-02-17
API rename prefixes "Item" for TextBinding / KeyBinding / ImageBinding

bchavez:
API rename prefixes "Item" for TextBinding / KeyBinding / ImageBinding on list-able controls


cwensley:
Work of wonder!  Thanks again!

Issue #271, opened 2015-02-18
Move GridView.CellClick/CellDoubleClick to Grid


cwensley:
These events are also needed for the TreeGridView.  It also looks like these are also only implemented in the WinForms platform.

PR #275, opened 2015-02-24
Implemented CellClick on Wpf ,Gtk and Mac.

BaseCrosser:
Implemented CellClick on Wpf and Gtk.

This implements the missing events that are mentioned #271.

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/GridHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/GridHandler.cs
@@ -125,10 +125,13 @@ namespace Eto.GtkSharp.Forms.Controls
 				case Grid.CellFormattingEvent:
 					SetupColumnEvents();
 					break;
+				case Grid.CellClickEvent:
+					Tree.ButtonPressEvent += OnTreeButtonPress;
+					break;
 				case Grid.CellDoubleClickEvent:
 					Tree.RowActivated += (sender, e) =>
 					{
-						var rowIndex = e.Path.Indices.Length > 0 ? e.Path.Indices[0] : -1;
+						var rowIndex = GetRowIndexOfPath(e.Path);
 						var columnIndex = GetColumnOfItem(e.Column);
 						var item = GetItem(e.Path);
 						var column = columnIndex == -1 ? null : Widget.Columns[columnIndex];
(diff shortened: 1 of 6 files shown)


BaseCrosser:
@cwensley I am looking at Mac now but I have never used MonoMac in the past.
I'll let you know in a couple of hours that I'll implement it on this PR or that I will create a new one for it if I figure it out eventually.

BaseCrosser:
@cwensley I have added CellClick in Mac. Could you check if I have done it right.
It works in the test application but I can't test Mac 64 bit version.

cwensley:
Hi looks really great!  Question - does Gtk and Wpf fire the event only when the mouse is clicked, or when the selection is changed also?  I haven't tested it, but it appears they might fire events incorrectly in that case.

BaseCrosser:
I'll investigate it tomorrow and see if I can change it if needed.

BaseCrosser:
@cwensley I fixed it. I think this is good now.

Source/Eto.Gtk/Forms/Controls/TreeGridViewHandler.cs
```diff
@@ -166,6 +166,44 @@ public override void AttachEvent(string id)
...
+				return;
+
+			int rowIndex = 0;
+			if (path.Indices.Length > 0)
+			{
+				for (int i = 0; i < path.Depth; i++)
+					rowIndex += path.Indices[i] + 1;
+
+				rowIndex--;
+			}
```
BaseCrosser: This part makes sure that the rowIndex is the right value.
I tested it in the test application and the row index is the same as the row index from the SelectionChanged event.

BaseCrosser:
Ok I'll change that. It happens always but I can change it in the original code actually.

BaseCrosser:
A nested tree shows its path like this 0:1:5 . If i take the first one (eg.Path.Indices[0]) it only returns the parent row. Now it returns the right row.

BaseCrosser:
I'm working on it. Don't have time right now. I'll check back in later.

Source/Eto.Gtk/Forms/Controls/GridHandler.cs
```diff
@@ -238,7 +268,27 @@ public int NumberOfColumns
...
 		public int GetColumnOfItem(Gtk.TreeViewColumn item)
 		{
-			return Widget.Columns.Select(r => r.Handler as GridColumnHandler).Select(r => r.Control).ToList().IndexOf(item);
+			for (int i = 0; i < Tree.Columns.Length; i++)
+				if (Tree.GetColumn(i) == item && i < Widget.Columns.Count)
+					return i;
+
+			return -1;
+		}
+
```
BaseCrosser: The original code did not work 100%.
It sometimes happened that the cast was unsuccesfull.
It was very irregular. Sometimes it worked, sometimes not.

I also tested the code below but it didn't work.
It seems that when casting `Widget.Columns[someColum,].Handler` to `GridColumnHandler` that sometimes (not every time) some data gets lost. One of the things that can get lost is the column title (it changes to `null`).
```C#
for (int i = 0; i < Widget.Columns.Count; i++)
{
    if ((Widget.Columns[i].Handler as GridColumnHandler).Control == item)
        return i;
}
return -1;
```
I tried to explain it the best I could but English is not my native language.

BaseCrosser:
I fixed it. The problem with GetColumnOfItem was my fault.
In my code i disposed the path and column and that was giving the trouble.
This is good to go now I think.

BaseCrosser:
@cwensley Is this good now?

Source/Eto.WinForms/Forms/Controls/GridHandler.cs
```diff
@@ -218,6 +218,12 @@ public override void AttachEvent(string id)
 				case Grid.CellClickEvent:
 					Control.CellClick += (sender, e) =>
 					{
+						if (e.RowIndex > -1)
```
cwensley: Will this will make the grid behave differently when the CellClick event is handled vs. not?

cwensley:
Hi, sorry it's taking a while for me to review - I've had a bunch of other things on my plate.  It is looking just about done!  Only that one last question about WinForms.

BaseCrosser:
No problem. About Winforms.. I didn't even think about that. No it doesn't. Do you want to keep that change? If so, I can move it so that it works without having cell click handled. If not, I can just remove it. It's up to you. Let me know and I'll change it.

cwensley:
Hm, I think removing it might be best, unless there's a specific reason for it to be there?

BaseCrosser:
Not really, I'll remove it.

BaseCrosser:
@cwensley Rebuild?

cwensley:
This looks awesome!  Thanks so much for your hard work.  I didn't realize that winforms did that with the GridView, though it doesn't seem to have the same issue with the TreeGridView.  There's probably something else going on there.

BaseCrosser:
I'm glad that I could help!

PR #278, opened 2015-02-27
* WinForms: Makes LinkButton text size behave similarly to Label

bchavez:
Basically, a copy over from LabelHandler in WinForms that makes LinkButton behave similarly when Text property is set after it's been added to a layout.


cwensley:
Great work!  Thanks!

PR #289, opened 2015-03-05
Implemented GetIterAtRow and GetPathAtRow on Gtk.

BaseCrosser:
I have added the ability to select a row with SelectRow(int row) in Gtk.
I have tested this against the Wpf test application and it behaves the same.

Also there was a bug when calling GetRowIndexOfPath from TreeGridView In Gtk but it is now fixed.


cwensley:
Beautiful, thank you for the contribution!

BaseCrosser:
You are welcome.

PR #290, opened 2015-03-06
Implemented SetSelectedRows in TreeGridView on Gtk.

BaseCrosser:
Implemented SetSelectedRows in TreeGridView on Gtk.
Also fixed a bug with SetSelectedRows in GridView on Gtk.

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/GridViewHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/GridViewHandler.cs
@@ -120,7 +120,7 @@ namespace Eto.GtkSharp.Forms.Controls
 							Tree.Selection.SelectIter(GetIterAtRow(start));
 						else
 							Tree.Selection.SelectRange(GetPathAtRow(start), GetPathAtRow(end));
-						start = end = -1;
+						start = end = row;
 					}
 				}
 				if (start != -1)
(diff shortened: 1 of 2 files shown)


Source/Eto.Gtk/Forms/Controls/GridViewHandler.cs
```diff
@@ -120,7 +120,7 @@ protected override void SetSelectedRows(IEnumerable<int> value)
 							Tree.Selection.SelectIter(GetIterAtRow(start));
 						else
 							Tree.Selection.SelectRange(GetPathAtRow(start), GetPathAtRow(end));
-						start = end = -1;
+						start = end = row;
```
BaseCrosser: When setting to -1 the current row gets skipped.

BaseCrosser:
@cwensley  Is it my fault that it is failing to build?

BaseCrosser:
@cwensley I think the problem is that it has the same version (build) number as one in the past that failed (the one that you changed the development version, about 4 months ago). I think it will be fixed when changing build number to 2.1.0.129 or higher.

BaseCrosser:
@cwensley Done building. Thanks for fixing the build server btw.

cwensley:
Very nice!  Thank you once again.  (;

PR #293, opened 2015-03-09
Fixed bug with cell click event in Wpf

BaseCrosser:
I've introduced a bug when adding cell click to wpf.
This fixes it.


cwensley:
Awesome (;

PR #296, opened 2015-03-13
Refined ProgressCell.

BaseCrosser:
Min value is forced to -1.
Max value is forced to 100.
Also I added that when the value is -1 it doesn't draw the ProgressCell.

### DIFF
--- a/Source/Eto.Gtk/Forms/Cells/ProgressCellHandler.cs
+++ b/Source/Eto.Gtk/Forms/Cells/ProgressCellHandler.cs
@@ -24,6 +24,9 @@ namespace Eto.GtkSharp.Forms.Cells
 
 			protected override void Render(Gdk.Drawable window, Gtk.Widget widget, Gdk.Rectangle background_area, Gdk.Rectangle cell_area, Gdk.Rectangle expose_area, Gtk.CellRendererState flags)
 			{
+				if (float.IsNaN((float)Handler.GetValueInternal(Item, Handler.ColumnIndex, Row).Val))
+					return;
+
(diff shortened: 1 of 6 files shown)


cwensley:
Hm, would it make sense to make it a nullable float?  if it is null, then it wouldn't show anything, and the percentage value is 0 - 1.  This will allow it to be more precise as well, and not limited to 100 steps.

BaseCrosser:
I'll look into it tommorow. Don't have much time today for it.

cwensley:
Awesome, no problem at all (;

BaseCrosser:
It won't work with the native progress bar on GTK# the way you suggested.
It can't handle a NULL value. What do you suggest? Keep it with Int values or draw it ourselfs on GTK# (sorta like DrawableCell)?

cwensley:
Could the float be translated to an int value or perhaps a float.NaN if it is null (for GTK specifically)?

BaseCrosser:
I got it to work on Gtk now. Thanks for the tip.
Mac also needed the float.NaN way.
I'll upload it in a minute or 5.

Quick question, is this right?
https://github.com/picoe/Eto/blob/develop/Source/Eto.Mac/Forms/FormHandler.cs#L70
Doesn't it need to be ```if (Control.IsVisible)```
I am asking this because OnShown doesn't fire on Mac nad when searching for the problem I came across this.

BaseCrosser:
I accidentaly deleted all my progress but I'm done now.
I think this is good to go now.

cwensley:
Great work again!  Thanks!

PR #310, opened 2015-03-27
Splitter startup sanity for FixedPanel != Panel1 fix #309 [Wpf,Swf,Gtk2]

firda-cze:
fix #309 

These changes will take care about initial position vs. inital (default, requested) size.
Fixes problems when FixedPanel != Panel1 and/or when not using designer (I prefer not to use designer except for simple dialogs or few user controls - but never for main form).

**NOTE:** Mac code not touched, will probably need similar changes. Gtk3 not tested.

Test code:

```C#
using System;
using System.Collections.Generic;
using System.Linq;

using Eto;
using Eto.Forms;
using Eto.Drawing;

using System.Threading;

namespace SplitterTests
{
	class MainForm : Form
	{
		[STAThread]
		static void Main()
		{
			new Application(Platforms.WinForms).Run(new MainForm());
		}

		MainForm()
		{
		#if DEBUG
			if(Platform.IsWinForms)
			{
				swfForm = this;
				swfApp = Application.Instance;
				var wpf = new Thread(delegate()
				{
					(wpfApp = new Application(Platforms.Wpf)).Run(
						wpfForm = new MainForm() {
						Location = new Point(60, 400) });
				});
				wpf.SetApartmentState(ApartmentState.STA);
				wpf.Start();
			}
			if (Platform.IsWpf)
			{
				var gtk = new Thread(delegate()
				{
					(gtkApp = new Application(Platforms.Gtk2)).Run(
						gtkForm = new MainForm() {
						Location = new Point(600, 60) });
				});
				gtk.SetApartmentState(ApartmentState.STA);
				gtk.Start();
			}
			Location = new Point(60, 60);
		#endif
			Size = new Size(400, 300);
		//	split.FixedPanel	= SplitterFixedPanel.Panel2;
			split.FixedPanel	= SplitterFixedPanel.None;
			split.Width			= 300;
			split.Position		= 100;
			split.Panel1		= area1;
			split.Panel2		= area2;
			Content = split;

			area1.SizeChanged += area_SizeChanged;
			area2.SizeChanged += area_SizeChanged;
		}

		Splitter split = new Splitter();
		TextArea area1 = new TextArea();
		TextArea area2 = new TextArea();

		void area_SizeChanged(object sender, EventArgs e)
		{
			area1.Text = string.Format(
				"W: {0}\r\nP: {1}\r\n({2}%)",
				area1.Width, split.Position,
				area1.Width*100/split.Width);
			area2.Text = string.Format(
[...]

PR #311, opened 2015-03-28
Splitter Sanity

### LINKED ISSUE #308: Adding a context menu to a Label doesn't work
using the following method:
```
label.MouseDown += (sender, e) => {
	if (e.Buttons != MouseButtons.Alternate)
		return;

	var button = new ButtonMenuItem () { Text = "Test" };
	var context = new ContextMenu (button);
	context.Show (sender as Label);
};
```
displays
image

Error is consistent on WPF, GTK2#, no menu appears at all on WinForms

### LINKED ISSUE #309: Splitter initial position with FixedPanel=None
I would expect to have the splitter sized 50:50 in following test code.
I will try to fix and pull it, just to let you know (chat is not perfect for code samples)

BTW: It will even crash with WPF if form.Width < splitter.Position. Negative GridLength calculated on line merged from my own last pull request... so... I will definitely look into it and fix it ;)

```C#
using System;
using System.Collections.Generic;
using System.Linq;

using Eto;
using Eto.Forms;
using Eto.Drawing;

namespace SplitterTests
{
	class MainForm : Form
	{
		[STAThread]
		static void Main()
		{
			new Application().Run(new MainForm());
			//new Application(Platforms.WinForms).Run(new MainForm());
			//new Application(Platforms.Direct2D).Run(new MainForm());
			//new Application(Platforms.Gtk2).Run(new MainForm());
		}

		MainForm()
		{
			Size = new Size(600, 400);
			var split = new Splitter() {
				FixedPanel	= SplitterFixedPanel.None,
				Width		= 1000,
				Position	= 500,
				Panel1 = new TextArea(),
				Panel2 = new TextArea()
			};
			Content = split;
		}
	}
}
```

firda-cze:
I finally got how WPF autosizing works - set_Position needs to update the control to make it work.
WPF/SWF and GTK#2 now support setting FixedPanel=None/Panel2 with Position and Width or Height to properly size it at startup. Autosizing checked and is working fine. WinForms needed one hack of EtoPanel to bypass system measurement for panel with one fully docked (filled) control (there is some cache-magic to blame from what I have seen in ILSpy). Original tests worked without it, but once I have added another splitter to the tests, it started to fail - fixed with that workaround.

list of squashed commints:

(merge) Fix MenuItemCollection.AddRange so that the items actually get added to the parent menu. Fixes #308
Splitter tests of initial resize (wpf: 1px error when fixing second)
Splitters improved, Wpf 1px error fixed, more tests
Hackish Swf autosizing fix for Dock=Fill in panel
Splitter tests
Minor change in tests to re-enable auto-size tests.
Wpf Splitter autosizing (removed return if Widget.Loaded in set_Position)

fix #309, replaces #310

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
@@ -114,6 +114,38 @@ namespace Eto.GtkSharp.Forms.Controls
 			}
 			if (position != null)
 				Control.Position = position.Value;
+
+			if (fixedPanel != SplitterFixedPanel.Panel1)
+				Control.SizeAllocated += Control_SizeAllocated;
+		}
+
+		void Control_SizeAllocated(object o, Gtk.SizeAllocatedArgs args)
+		{
+			Control.SizeAllocated -= Control_SizeAllocated;
+			if (position == null)
+				return;
+			if (orientation == SplitterOrientation.Horizontal)
+			{
+				int width = PreferredSize.Width;
+				if (width <= 0)
+					return;
+				if (fixedPanel == SplitterFixedPanel.Panel2)
+					Control.Position = Math.Max(0,
+						position.Value + args.Allocation.Width - width);
+				else
+					Control.Position = position.Value * args.Allocation.Width / width;
+			}
+			else
+			{
+				int height = PreferredSize.Height;
+				if (height <= 0)
+					return;
+				if (fixedPanel == SplitterFixedPanel.Panel2)
+					Control.Position = Math.Max(0,
+						position.Value + args.Allocation.Height - height);
+				else
+					Control.Position = position.Value * args.Allocation.Height / height;
+			}
 		}
 
 		static Gtk.Widget EmptyContainer()
--- a/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
@@ -10,7 +10,9 @@ namespace Eto.GtkSharp.Forms.Controls
 		Control panel2;
 		SplitterOrientation orientation;
 		SplitterFixedPanel fixedPanel;
+		SplitterPositionMode mode;
 		int? position;
+		bool created;
 
 		class EtoHPaned : Gtk.HPaned
 		{
@@ -57,8 +59,65 @@ namespace Eto.GtkSharp.Forms.Controls
 
 		public int Position
 		{
-			get { return Control.Position; }
-			set { position = Control.Position = value; }
+			get
+			{
+				switch (mode)
+				{
+					default:
+						return Control.Position;
+					case SplitterPositionMode.Far:
+						return Math.Max(0, (Orientation == SplitterOrientation.Horizontal ?
+							Control.Allocation.Width : Control.Allocation.Height) - Control.Position - SplitterWidth);
+					case SplitterPositionMode.Percent:
+						var width = (Orientation == SplitterOrientation.Horizontal ?
+							Control.Allocation.Width : Control.Allocation.Height) - SplitterWidth;
+						return width <= 0 ? 0 : Control.Position * 100 / width;
+				}
+			}
+			set
+			{
+				position = Control.Position = value;
+			}
+		}
+
+		int GetAvailableSize()
+		{
+			return GetAvailableSize(!created);
+		}
+		int GetAvailableSize(bool desired)
+		{
+			if (desired)
+			{
+				var size = PreferredSize;
+				var pick = Orientation == SplitterOrientation.Horizontal ?
+					size.Width : size.Height;
+				if (pick >= 0)
+					return pick - SplitterWidth;
+			}
+			return (Orientation == SplitterOrientation.Horizontal ?
+				Control.Allocation.Width : Control.Allocation.Height) - SplitterWidth;
+		}
+
+		void SetPosition(int pos, SplitterPositionMode mode)
+		{
+			int size = GetAvailableSize(false);
+			if (mode == SplitterPositionMode.Far)
+				pos = size - pos;
+			else if (mode == SplitterPositionMode.Percent)
+				pos = (1 + 2 * size * pos) / 200;
+			Control.Position = Math.Max(0, Math.Min(size, pos));
+		}
+
+		public SplitterPositionMode PositionMode
+		{
+			get { return mode; }
+			set { mode = value; }
+		}
+
+		public int SplitterWidth
+		{
+			get { return 5; /* just u guess :( */ }
+			set { /* unfortunatelly I cannot see binding for gtk_set_gutter_size in Gtk# */ }
 		}
 
 		public SplitterFixedPanel FixedPanel
@@ -69,7 +128,8 @@ namespace Eto.GtkSharp.Forms.Controls
 				if (fixedPanel != value)
 				{
 					fixedPanel = value;
-					Create();
+					((Gtk.Paned.PanedChild)Control[Control.Child1]).Resize = value != SplitterFixedPanel.Panel1;
+					((Gtk.Paned.PanedChild)Control[Control.Child2]).Resize = value != SplitterFixedPanel.Panel2;
 				}
 			}
 		}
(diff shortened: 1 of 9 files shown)


Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
```diff
@@ -114,6 +114,38 @@ void Create()
 			}
 			if (position != null)
 				Control.Position = position.Value;
+
+			if (fixedPanel != SplitterFixedPanel.Panel1)
+				Control.SizeAllocated += Control_SizeAllocated;
```
cwensley: This may be called more than once, causing the event to be subscribed multiple times

Source/Eto/Forms/Menu/MenuItemCollection.cs
```diff
@@ -134,15 +134,9 @@ public void AddSeparator(int order = 0)
 		/// <param name="items">Items to add.</param>
 		public void AddRange(IEnumerable<MenuItem> items)
 		{
-			var list = Items as List<MenuItem>;
```
cwensley: These changes should already be fixed in the develop branch, not sure why they're showing on the pull.  Perhaps you can rebase then force push to the branch?

cwensley:
Looking really good!  Thanks for taking the time to make these fixes.  I think the only real issue is the subscription of the event multiple times in gtk, which should hopefully be an easy fix.

firda-cze:
1. The event cannot be subscribed more then once, because Create() always crate new. The old one is discarded together with old subscribtion.

```C#
		void Create()
		{
			Gtk.Paned old = Control;
			if (orientation == SplitterOrientation.Horizontal)
				Control = new EtoHPaned();
			else
				Control = new EtoVPaned();
```
  ...but there still may be problem with readding (the event unsubscribe itself upon first execution - so again, no, cannot be subscribed more than once).

2. Some git mess, I have merged your change to my branch and then did some rebase (which was probably wrong, I should have rebsed to your version). Still learning ;)

3. I am currently working on even better version, see my work branch (SplitterPositionMode: Near, Far, Percent). ...but I will need help with Mac version (I have no way to test it). Currently I have working versions for WinForms and WPF, now working on GTK. A plan to add more tests (especially involing removing and readding full splitter as that can cause problems - e.g. with that GTK SizeAllocated event).

Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
firda-cze: Control is created anew on each call of Create() method.

Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
```diff
@@ -114,6 +114,38 @@ void Create()
...
+		}
+
+		void Control_SizeAllocated(object o, Gtk.SizeAllocatedArgs args)
+		{
+			Control.SizeAllocated -= Control_SizeAllocated;
```
firda-cze: ...but this may be a problem, I am not sure how would that behave when removed from parent and then added again or to different widget

Source/Eto/Forms/Menu/MenuItemCollection.cs
firda-cze: I have probably messed some merge/rebase (I have merged your develop to mine).

Source/Eto.Wpf/Forms/Controls/SplitterHandler.cs
```diff
@@ -6,18 +6,20 @@
 using swc = System.Windows.Controls;
 using Eto.Drawing;
 
+using System.Diagnostics;
```
firda-cze: Unnecessary

Source/Eto.Wpf/Forms/Controls/SplitterHandler.cs
```diff
@@ -51,35 +53,62 @@ public SplitterHandler()
...
+				if (fixedPanel == SplitterFixedPanel.Panel2)
+				{
+					var prefer = panel2.GetPreferredSize(Conversions.PositiveInfinitySize);
+					Position = Math.Max(0, (int)Math.Round(
+						size - splitterSize - (horiz ? prefer.Width : prefer.Height)));
+					return;
 				}
-				else if (position != null)
+				var sone = panel1.GetPreferredSize(Conversions.PositiveInfinitySize);
+				var stwo = panel1.GetPreferredSize(Conversions.PositiveInfinitySize);
```
firda-cze: bug, should be panel2

Source/Eto.WinForms/Forms/Controls/SplitterHandler.cs
```diff
@@ -202,7 +229,16 @@ void SetInitialPosition()
 							pos = Control.Width - size2.Width;
 						else
 							pos = Control.Height - size2.Height;
-						SetPosition(pos);
+						SetPosition(Math.Max(Control.Panel1MinSize, pos - Control.SplitterWidth));
+						break;
+					default:
+						var sone = panel1.GetPreferredSize();
+						var stwo = panel1.GetPreferredSize();
```
firda-cze: bug, should be panel2

firda-cze:
Let me know which version would you like
...and if you are willing to finish the Mac version.
Current is here: https://github.com/firda-cze/Eto/tree/work
(now compilable, WinForms/WPF and GTK finished, Mac stub only, I plan to add more tests, especially for removing and readding widgets/splitters).

cwensley:
Hi @firda-cze, 

Sorry I didn't realize you were waiting for me.  I am thinking that the new SplitterPositionMode is certainly explicit in how to use the position, but it also will break all existing code (which is not ideal).

Is there a way that you can think to keep it backwards compatible while still providing the extended functionality?

Also, in terms of percent, it would be nice if it were 0-1, instead of 0-100, in keeping with how other API's in the framework are designed.

cwensley:
Here's an idea, let me know what you think:

The SplitterPositionMode may not be necessary, as it can be inferred based on the FixedPanel.  We can add a separate property (e.g. RelativePosition or some better name) which would be like 'Near' if FixedPanel = Panel1, 'Far' when Panel2, and 'Percent' when None.

Thoughts?  This does limit the functionality a bit so that if you wanted to set the position from the Right/Bottom but have FixedPanel set to Panel1, but that is highly unlikely and they can use the traditional Width and Position method, or set the size of the non-fixed panel explicitly for that purpose.

firda-cze:
It all depends on how much backward-compatible you want to be.

The `SplitterPositionMode` is there exactly **not to break existing code**, because default would be `Near`, which is the compatible behaviour. I would as well prefer `double 0.0 .. 1.0` instead of percent, but you have it `int Position`.

I was thinking about adding `SplitterPositionMode.Auto` that would behave exactly as you have described (and static flag to make it default behaviour if requested at the start of your application - not to break any old code).

It was great opportunity to learn about internals of the toolkit, trying to enhance the splitters without breaking anything, but I would personally change `Position` to return `double` instead of `int` and maybe forget about native implementations and create one multi-plaftorm on top of Eto.

Part of my own WinForms Splitter for inspiration (note, that it can distribute the remaining space in unfixed mode - once you substract minimal sizes and splitter width - that scales nicely):

```C#
protected override void OnLayout( LayoutEventArgs e )
{
	if( p1.Hidden ) // first panel is hidden => second takes all the place
		p2.SetBounds( 0, 0, Width, Height );
	else if( p2.Hidden ) // second panel is hidden => first takes all the place
		p1.SetBounds( 0, 0, Width, Height );
	else if( horiz ) // horizontal in WinForms speaks about the splitter/gutter (opposite from Eto)
	{
		if( p2.Fixed ) // when in fixed mode, ratio is position in pixels (like Near/Far mode)
		{ // but it reads "extra width/height" on top of minimal and is zero if minimal = cannot shrink anymore
			int h = Math.Min( Height, p2.MinSize + (int)Math.Round( ratio ) );
			p1.SetBounds( 0, 0, Width, Math.Max( 0, Height-h-swidth ) );
			p2.SetBounds( 0, Height-h, Width, h );
		}
		else
		{ // in unfixed mode, it is the ratio (0..1) for remaining space distribution (default 0.5 = 50:50)
			int h = Math.Min( Height, p1.MinSize + (int)Math.Round(
				p1.Fixed ? ratio : (Height-swidth-p1.MinSize-p2.MinSize)*ratio ) );
			p1.SetBounds( 0, 0, Width, h );
			h += swidth;
			p2.SetBounds( 0, h, Width, Math.Max( 0, Height-h ) );
		}
	}
	else if( p2.Fixed )
	{
		int w = Math.Min( Width, p2.MinSize + (int)Math.Round( ratio ) );
		p1.SetBounds( 0, 0, Math.Max( 0, Width-w-swidth ), Height );
		p2.SetBounds( Width-w, 0, w, Height );
	}
	else
	{
		int w = Math.Min( Width, p1.MinSize + (int)Math.Round(
			p1.Fixed ? ratio : (Width-swidth-p1.MinSize-p2.MinSize)*ratio ) );
		p1.SetBounds( 0, 0, w, Height );
		w += swidth;
		p2.SetBounds( w, 0, Math.Max( 0, Width-w ), Height );
	}
	base.OnLayout( e );
}
protected override void OnMouseMove( MouseEventArgs e )
{
	if( (e.Button & MouseButtons.Left) != 0 && !p1.Hidden && !p2.Hidden )
	{
		double w = horiz
			? p1.Fixed ? e.Y-(swidth>>1)-p1.MinSize
			: p2.Fixed ? Height-e.Y-(swidth>>1)-p1.MinSize
			: (e.Y-(swidth>>1)-p1.MinSize)/(double)(Height-swidth-p1.MinSize-p2.MinSize)
			: p1.Fixed ? e.X-(swidth>>1)-p1.MinSize
			: p2.Fixed ? Width-e.X-(swidth>>1)-p1.MinSize
			: (e.X-(swidth>>1)-p1.MinSize)/(double)(Width-swidth-p1.MinSize-p2.MinSize);
		if( w != ratio )
		{
			Weight = w;
			Events.Invoke( EVT_SplitterMoved, this, EventArgs.Empty );
		}
	}
	base.OnMouseMove( e );
}
```

firda-cze:
I like the idea of `double RelativePositon`, that would keep `int Position` for backward compatibility (and `Near` mode) and the new one for sanity (`Auto` mode = `Near/Far/Ratio` according to `FixedPanel`).

Try to think about the universal (without native) implementation and about additional size for each panel (for nicer ratio/unfixed/none distribution - I am using this mode in my own IDE app with 120px min and 0.15 additional scaling for left panel).

BTW: This is a screenshot from application I plan to port to Eto (darker style selected). You can see a lot of splitters there.... and my custom PropertyGrid, that I plan to implement using Eto.
ide

firda-cze:
This is my current attempt for RelativePosition:
https://github.com/firda-cze/Eto/commit/8ca55aeeea7d538c8587d76ad894f10603b003a8?ts=4

Let me know which way to go (all native handlers or one universal splitter for all platforms written directly in Eto - that should be possible and easiest way ... can we do hybrid? I mean some default handler that you can override for some platforms if you wish but leave ot on the default for most of them). It should not be a big problem to write GTK and WPF implementation, but I cannot do Mac.

cwensley:
Keeping the native handlers is probably the way to go for now, to allow for nuances between platforms, and to use the standard controls where possible.  

If you want, you can add a custom-implemented Splitter in Eto.Forms.ThemedControls like there is for the Spinner, where people would have to explicitly add it to their platform instance like:

myPlatform.Add<Splitter.IHandler>(() => new ThemedSplitterHandler());

Cheers!
Curtis.

firda-cze:
One question about `ThemedSplitterHandler`: How can I provide auto-sizing?

cwensley:
There's no way to do auto sizing with the Eto api's directly at the moment (but I am working on this).  `ThemedContainerHandler` is used when your control contains other controls, which is appropriate for the Splitter.

firda-cze:
OK, TableLayout looks like good base class (I am currently testing it under Direct2D and DEBUG, but think that it could be made the default, because WinForms SplitterContainer is not moving fluently - selecting D2D may be fine to select smoother implementation).

I have some notes about auto-sizing and WinForms GetPreferredSize:
I think that anything <= 0 should be treated as "say minimal" that you can freely substract padding and such without always using `Math.Max(0, ...)`
and anything >= `short.MaxValue` (or some constant) as "say preferred" (again about substracting)
and stop using `-1` for auto-sizing but rather use some `int.MinValue` or `short.MinValue` (but that could cause compatibility problems).

It is all about math during passing the *constraint* (preferredSize) to inner containers while doing some math involving padding/margin and such. It would be much easier if the final control doing the true math would check the constraint to figure-out what is what instead of using `Math.Max` and `if (blah == int.MaxValue) ...`. Got it? What do you think?

firda-cze:
Can you look at this diff?
https://github.com/picoe/Eto/compare/develop...firda-cze:Splitter.RelativePosition?ts=4
I will probably create new pull request from this as it got changed a lot and I want to delete my "develop" branch.

There is still some problem with WPF auto-sizing (although it passes unit tests)... and I would like to understand, why is the `replay` working only for WinForms. ThemedSplitter is working, but not auto-sizing. Mac untouched.

firda-cze:
I have done and tested what I could. Let me know how to proceed.
*This pull request is outdated, I will create new one or rewrite my develop branch with Splitter.RelativePosition branch*

WinForms should be perfect - I am developing for WinForms for quite some time.
WPF looks good, but I still don't know all the corner-cases and am not sure about the events (will check later and possibly add more unit tests).
GTK2 should work fine, it is mostly based on WinForms implementation. GTK3 has different sizing mechanism, so, I guessed it (cannot curretnly test, but I will probably create some Linux VM in near future). SplitterWidth is guessed as well. I am not so familiar with GTK.
**There is a stub for Mac, that should not break existing code, but new features are up to you.**
ThemedSplitter is unfinished, tested under WinForms/D2D, but is not auto-sizing.

I'd like some feedback about the new unit tests (`Shown`) and why is it only working with WinForms when I set `replay: true`. It was designed to replay the test again, once the form is on the screen - remove and readd and retest the conrol. WPF/GTK passed the tests without `replay` (initialzed with the form, tested once shown).

cwensley:
Hi @firda-cze, that's great!  A stub for mac is perfect, I can take a look at implementing it.

As for the unit test problem, I'm not sure what is going wrong there.. quite possibly because the layout of the controls aren't done immediately.. You may have to do an Application.Instance.AsyncInvoke() to test it the second time.

firda-cze:
I did try `AsyncInvoke` for GTK and it did not help. Yes, it seems to be problem with layout schedule... any multiplatform approach? How can I ensure in Eto that all layout-logic is done?

What about this pull, should I rewrite my develop branch or create new pull request? ...I'd prefer creating new, because the branch has good name and should look better in git history.

BTW: Thanks for the hint to use SourceTree - **Ancestor Order** is just wonderfull, no need for rebase anymore.

Issue #319, opened 2015-04-02
Add a feature to set the startup location of dialogs


Jojatekok:
I would like to set a Dialog's position to the center of its parent Window, but it seems that I cannot even get the parent Window of Dialogs...

_Related: http://stackoverflow.com/questions/4306593/wpf-how-to-set-a-dialog-position-to-show-at-the-center-of-the-application_

Panzerfather:
Generally the window is set up to center position if you open it like this:

MyDialog.ShowModal(this); <= where "MyDialog" is the Dialog to open and "this" is the parent control.

You can then move it via the Location property.

BUT right now, Eto isn't setting up the Parent property of the control when runnig ShowModal-Functions.

Quick solution is to change following functions in the Eto.sln:

Forms\Dialog.cs
=======================
/// <summary>
/// Shows the dialog modally, blocking the current thread until it is closed.
/// </summary>
/// <remarks>
/// The <paramref name="parent"/> specifies the control on the window that will be blocked from user input until
/// the dialog is closed.
/// </remarks>
/// <param name="parent">Parent control that is showing the form</param>
public void ShowModal(Control parent = null)
{
	base.Parent = parent as Container;

	bool loaded = Loaded;
	if (!loaded)
	{
		OnPreLoad(EventArgs.Empty);
		OnLoad(EventArgs.Empty);
		OnDataContextChanged(EventArgs.Empty);
		OnLoadComplete(EventArgs.Empty);
	}

	Application.Instance.AddWindow(this);

	Handler.ShowModal(parent);
}

/// <summary>
/// Shows the dialog modally asynchronously
/// </summary>
/// <remarks>
/// The <paramref name="parent"/> specifies the control on the window that will be blocked from user input until
/// the dialog is closed.
/// </remarks>
/// <param name="parent">Parent control that is showing the form</param>
public Task ShowModalAsync(Control parent = null)
{
	base.Parent = parent as Container;

	bool loaded = Loaded;
	if (!loaded)
	{
		OnPreLoad(EventArgs.Empty);
		OnLoad(EventArgs.Empty);
		OnDataContextChanged(EventArgs.Empty);
		OnLoadComplete(EventArgs.Empty);
	}

	return Handler.ShowModalAsync(parent);
}

Jojatekok:
Thank you! I just wanted to use the parent Window's properties :)

Jojatekok:
This issue shouldn't be closed until it's pulled to the main repository

[...]

PR #321, opened 2015-04-02
Add Window.Owner. Patch for Issue #319

Panzerfather:
Setup parent when using Dialog.ShowModal

### DIFF
--- a/Source/Eto.Gtk/Forms/GtkWindow.cs
+++ b/Source/Eto.Gtk/Forms/GtkWindow.cs
@@ -249,7 +249,7 @@ namespace Eto.GtkSharp.Forms
 					Control.DeleteEvent += Connector.HandleDeleteEvent;
 					break;
 				case Eto.Forms.Control.ShownEvent:
-					Control.Shown += Connector.HandleShown;
+					Control.Shown += Connector.HandleShownEvent;
 					break;
 				case Window.WindowStateChangedEvent:
 					Connector.OldState = WindowState;
@@ -287,7 +287,7 @@ namespace Eto.GtkSharp.Forms
 				args.RetVal = !Handler.CloseWindow();
 			}
 
-			public void HandleShown(object sender, EventArgs e)
+			public void HandleShownEvent(object sender, EventArgs e)
 			{
 				Handler.Callback.OnShown(Handler.Widget, EventArgs.Empty);
 			}
--- a/Source/Eto/Forms/Form.cs
+++ b/Source/Eto/Forms/Form.cs
@@ -31,7 +31,7 @@ namespace Eto.Forms
 		/// Show the form
 		/// </summary>
 		public void Show()
-		{ 
+		{
 			var loaded = Loaded;
 			if (!loaded)
 			{
--- a/Source/Eto/Forms/Window.cs
+++ b/Source/Eto/Forms/Window.cs
@@ -118,6 +118,29 @@ namespace Eto.Forms
 		}
 
 		/// <summary>
+		/// Identifier for handlers when attaching the <see cref="OwnerChanged"/> event.
+		/// </summary>
+		private const string OwnerChangedEvent = "Window.OwnerChanged";
+
+		/// <summary>
+		/// Occurs when the <see cref="Owner"/> is changed.
+		/// </summary>
+		public event EventHandler<EventArgs> OwnerChanged
+		{
+			add { Properties.AddEvent(OwnerChangedEvent, value); }
+			remove { Properties.RemoveEvent(OwnerChangedEvent, value); }
+		}
+
+		/// <summary>
+		/// Raises the <see cref="OwnerChanged"/> event.
+		/// </summary>
+		/// <param name="e">Event arguments</param>
+		protected virtual void OnOwnerChanged(EventArgs e)
+		{
+			Properties.TriggerEvent(OwnerChangedEvent, this, e);
+		}
+
+		/// <summary>
 		/// Identifier for handlers when attaching the <see cref="WindowStateChanged"/> event.
 		/// </summary>
 		public const string WindowStateChangedEvent = "Window.WindowStateChanged";
@@ -139,6 +162,12 @@ namespace Eto.Forms
 		{
 			Properties.TriggerEvent(WindowStateChangedEvent, this, e);
 		}
+
+		#endregion
+
+		#region Properties
+
+		Window owner = null;
 
 		#endregion
 
@@ -246,6 +275,23 @@ namespace Eto.Forms
 		public virtual void Close()
 		{
 			Handler.Close();
+		}
+
+		/// <summary>
+		/// Gets or sets the owner of this window.
+		/// </summary>
+		/// <value>The owner of this window.</value>
+		public Window Owner
+		{
+			get { return this.owner; }
+			set
+			{
+				if (this.owner != value)
+				{
+					this.owner = value;
+					this.OnOwnerChanged(EventArgs.Empty);
+				}
+			}
 		}
 
 		/// <summary>
@@ -383,7 +429,7 @@ namespace Eto.Forms
 		}
 
 		/// <summary>
-        /// Sets <see cref="WindowState"/> to <see cref="Eto.Forms.WindowState.Minimized"/>
+		/// Sets <see cref="WindowState"/> to <see cref="Eto.Forms.WindowState.Minimized"/>
 		/// </summary>
 		public void Minimize()
 		{
--- a/Source/Eto.Test/Eto.Test/Sections/Behaviors/WindowsSection.cs
+++ b/Source/Eto.Test/Eto.Test/Sections/Behaviors/WindowsSection.cs
@@ -240,8 +240,28 @@ namespace Eto.Test.Sections.Behaviors
 			{
 				var dialog = new Dialog();
 				child = dialog;
-				show = () => dialog.ShowModal(this);
+				show = () => dialog.ShowModal();
 			}
+
+			var layout = new DynamicLayout();
+			layout.Add(null);
+			layout.AddCentered(TestChangeSizeButton());
+			layout.AddCentered(TestChangeClientSizeButton());
+			layout.AddCentered(SendToBackButton());
+			layout.AddCentered(CreateCancelClose());
+			layout.AddCentered(CloseButton());
+			layout.Add(null);
+			child.Content = layout;
+
+			child.OwnerChanged += child_OwnerChanged;
+			child.WindowStateChanged += child_WindowStateChanged;
+			child.Closed += child_Closed;
+			child.Closing += child_Closing;
+			child.Shown += child_Shown;
+			child.GotFocus += child_GotFocus;
+			child.LostFocus += child_LostFocus;
+			child.LocationChanged += child_LocationChanged;
+			child.SizeChanged += child_SizeChanged;
 
 			child.Title = "Child Window";
 			child.WindowStyle = styleCombo.SelectedValue;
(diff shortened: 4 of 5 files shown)


cwensley:
Thanks for the pull request!  It is really appreciated.  A few questions:

What happens if you pass a Button instance to ShowModal?  The parent will still be null.  

Also, using the existing Parent property is probably not a good idea, as it is used to tell which control is the containing parent in the UI hierarchy, which is a different purpose for this.. It might be better to have an Owner property (of type Window) added to the Dialog class that would be set to the control.ParentWindow when ShowModal is called.

Thoughts?

Panzerfather:
I'm not quite sure what you mean with "what happens if you pass a Button instance to ShowModal?". Eto.Button doesn't inherit a ShowModal method.

The former patch was using the as-cast, so if the parent is null, it will be set to null. No error will occur. It doesn't matter which control inherits the method.

But I agree with your idea of integrating an owner property, so I've done this in my latest patch. The patches for Android, IOS and WinRT are untested, but should be functional.

cwensley:
@Panzerfather, thanks for the feedback!  I'll try to explain what I am thinking regarding those..

The problem with making the Owner property writable is that it is not doing anything when you set it.  Say for example you create a Dialog, set it's Owner, then call ShowModal(), it won't appropriately set the owner for each platform, whereas as a consumer of the API I would expect it to.

If we want to make it writable, then the handler implementations should use the property when showing the dialog instead of using the Control parameter (e.g. the parent parameter should then be removed from Dialog.IHandler.ShowModal and ShowModalAsync).

As for moving this to Dialog - Wpf and WinForms don't have a separation of Dialog (modal) vs. Form (modeless) - this property won't do anything on a Form, so it should probably not be available there.  You could still do as you describe by closing all non-modal windows by closing all Form windows and skipping Dialog windows.

Panzerfather:
@cwensley, thanks for the explanation. I removed the Handler-Code in the latest commit and added a property and an event to the Window class. I also made small changes to the Dialog and Form classes

Let me show and explain why it makes sence to use the Window class and not only the Dialog class for implementation:

application

As you can see in the screenshot there is a Main-Form which you should imagine as a Control-Window. From this form you can run and show with multiple windows as usual.
But the main difference is that if the Owner-Property is setup in the Window-Class the Sub-[Window/-Dialog/Form] can interact with the Main-Form without having to maintain e.g. your own list of opened windows.
e.g. the "Close all windows which have me as owner"-Button have just following code:
```C#
protected void OnButtonCloseAllClick(object sender, EventArgs e)
{
	// Have to be list or equivalent to avoid enumerable changed error
	List<Window> lstWindows = Application.Instance.Windows.Where(w => w.Owner == this).ToList();

	foreach (Window winOpen in lstWindows)
	{

		winOpen.Close();

	}
}
```

And it closes all Windows which have the Main-Form as owner. All others will remain open. You could also easily iterate through a Sub-SubWindow with this approach, to close them, update them, whatever.

This example can also be done when using .NET WinForms and WPF and set the Owner-Property manually.

As a good multiplatform example, Gimp is using such a approach if you are not running it in SingleWindow-Mode. But they have to maintain there own Handling-Code, which isn't fail-safe like this one.

Source/Eto/Forms/Window.cs
```diff
@@ -118,6 +118,30 @@ protected virtual void OnLocationChanged(EventArgs e)
...
+		EventHandler<EventArgs> OwnerChangedEvent;
+
+		/// <summary>
+		/// Occurs when the <see cref="Owner"/> is changed.
+		/// </summary>
+		public event EventHandler<EventArgs> OwnerChanged
+		{
+			add { OwnerChangedEvent += value; }
+			remove { OwnerChangedEvent -= value; }
+		}
```
cwensley: This should be changed to use the Properties.AddEvent() and Properties.RemoveEvent() so as to remove memory overhead when not handled

Source/Eto/Forms/Window.cs
```diff
@@ -118,6 +118,30 @@ protected virtual void OnLocationChanged(EventArgs e)
...
+
+		/// <summary>
+		/// Raises the <see cref="OwnerChanged"/> event.
+		/// </summary>
+		/// <param name="e">Event arguments</param>
+		protected virtual void OnOwnerChanged(EventArgs e)
+		{
+			if (OwnerChangedEvent != null)
```
cwensley: Should be `Properties.TriggerEvent(OwnerChangedEvent, this, e)`

Source/Eto/Forms/Form.cs
```diff
@@ -30,8 +30,10 @@ public Form(IHandler handler)
 		/// <summary>
 		/// Show the form
 		/// </summary>
-		public void Show()
-		{ 
+		public void Show(Window owner = null)
```
cwensley: I was thinking of reverting this in favour of using the Owner property directly.. Same with Dialog.ShowModal/Async() to deprecate the existing method(s) and add overloads without the owner parameter..  I think that provides a much cleaner api.. e.g. 
```c#
var dlg = new Dialog { Owner = this };
dlg.ShowModal();
```

Thoughts?

cwensley:
This looks like it needs a rebase.. the diffs show all the changes that are unrelated to the pull

Panzerfather:
Rebased it.

Source/Eto.Gtk/Forms/GtkWindow.cs
```diff
@@ -249,7 +249,10 @@ public override void AttachEvent(string id)
 					Control.DeleteEvent += Connector.HandleDeleteEvent;
 					break;
 				case Eto.Forms.Control.ShownEvent:
-					Control.Shown += Connector.HandleShown;
+					Control.Shown += Connector.HandleShownEvent;
+					break;
+				case Window.OwnerChangedEvent:
```
cwensley: This shouldn't be necessary, this doesn't need to be a HandlerEvent..  I believe this will cause the event to be called more than once, since you are already calling it when setting the Window.Owner property..

cwensley:
I think for the OwnerChanged event, either it needs to be handled in each of the platform handlers, or have it trigger when the property is set in Eto.Forms.Window.Owner.. it should probably not be called by both..

Source/Eto.Gtk/Forms/GtkWindow.cs
Panzerfather: If it isn't handled in the Platform-Handling, there's an exception of a missing event. That was the reason I didn't used the Properties-Event handling in f5d7b493c14c8f7b60ce2e358ec9982c8fa64dd3

Panzerfather:
Cleaned up code and removed platform handling.

Source/Eto.Gtk/Forms/GtkWindow.cs
Panzerfather: Changed Property-Event declaration to remove platform handling code in the next commit 7f696f6f9951299fa9c838c652e94b19b502b498.

Source/Eto.Gtk/Forms/GtkWindow.cs
```diff
@@ -287,11 +287,16 @@ public void HandleDeleteEvent(object o, Gtk.DeleteEventArgs args)
 				args.RetVal = !Handler.CloseWindow();
 			}
 
-			public void HandleShown(object sender, EventArgs e)
+			public void HandleShownEvent(object sender, EventArgs e)
 			{
 				Handler.Callback.OnShown(Handler.Widget, EventArgs.Empty);
 			}
 
+			public void HandleOwnerEvent(object sender, EventArgs e)
```
cwensley: This should be removed now that we aren't using the handler event

Source/Eto.Mac/Forms/MacWindow.cs
```diff
@@ -289,6 +289,14 @@ static bool HandleWindowShouldClose(NSObject sender)
 			return !args.Cancel;
 		}
 
+		static void HandleOwnerChanged(object sender, EventArgs e)
```
cwensley: Should be removed also

Source/Eto.Mac/Forms/MacWindow.cs
```diff
@@ -373,6 +384,18 @@ public override void AttachEvent(string id)
 						Control.WillMove += HandleWillMove;
 					}
 					break;
+				case Window.OwnerChangedEvent:
```
cwensley: More removal.. (;

cwensley:
Some of the platform handling still seems to be there.. could you take a look?

Panzerfather:
Pushed missing cleanup, now all platform code should be removed.

cwensley:
Definitely getting there!   I'm still seeing a wee bit of platform code lingering.. in the Callback/ICallback, the static constructor of Window, and the Window.OwnerChangedEvent id should be private (only should be public for handler-based events).

Panzerfather:
This change can't be integrated. For a Callback the private keyword isn't allowed and would create compiler errors.
Also the static Window() method can't have an keyword, keyword modifiers aren't allowed for static constructors. This would also create compiler errors.

Panzerfather:
Made the OwnerChangedEvent constant private and added a an check to Eto.Test. Restructuring the child changing a little bit to ensure that all events will be fired. Event-Callback overloads were made later in code than the changes of the Window/Dialog itself.

Source/Eto.WinRT/Forms/WpfWindow.cs
```diff
@@ -109,6 +109,9 @@ public override void AttachEvent (string id)
 					}
 				};
 				break;
+			case Window.OwnerChangedEvent:
```
cwensley: This is still some handler code that should be removed

Source/Eto/Forms/Window.cs
```diff
@@ -142,11 +165,18 @@ protected virtual void OnWindowStateChanged(EventArgs e)
...
+		Window owner = null;
+
+		#endregion
+
 		static Window()
 		{
 			EventLookup.Register<Window>(c => c.OnClosed(null), ClosedEvent);
 			EventLookup.Register<Window>(c => c.OnClosing(null), ClosingEvent);
 			EventLookup.Register<Window>(c => c.OnLocationChanged(null), LocationChangedEvent);
+			EventLookup.Register<Window>(c => c.OnOwnerChanged(null), OwnerChangedEvent);
```
cwensley: This is only needed for handler-based events

Source/Eto/Forms/Window.cs
```diff
@@ -451,6 +502,10 @@ public new interface ICallback : Panel.ICallback
 			/// </summary>
 			void OnLocationChanged(Window widget, EventArgs e);
 			/// <summary>
+			/// Raises the owner changed event.
+			/// </summary>
+			void OnOwnerChanged(Window widget, EventArgs e);
```
cwensley: This is also not necessary, as the handlers aren't triggering this event.

Panzerfather:
Made the changes, sorry for the forgotten removal of the WinRT platform code. It's a shame that VS 2013 can't handle/open WinRT-Projects on none Windows 8.x platforms! So they aren't even searchable for code-fragments. Maybe therefor I won't change code for this platform anymore. Making to much trouble when adding/removing.

cwensley:
No problem at all!  It happens to everyone (including myself) from time to time.  I believe Win 10 will be free upgrade for Win 7 so at least you'll have a way to do it in the future.

A squash would be really nice on this pull as well if that isn't too much trouble?

Source/Eto/Forms/Window.cs
```diff
@@ -483,6 +529,13 @@ public void OnLocationChanged(Window widget, EventArgs e)
 				widget.Platform.Invoke(() => widget.OnLocationChanged(e));
 			}
 			/// <summary>
+			/// Raises the owner changed event.
+			/// </summary>
+			public void OnOwnerChanged(Window widget, EventArgs e)
```
cwensley: One last bit to remove.. (;

Panzerfather: Done it, working on a squash right now.

Panzerfather:
Squashed it.

cwensley:
Beautiful!  Nice and clean, thank you so much!  I like how you picked up on the Obsolete convention for specifying the version it was deprecated (;

PR #328, opened 2015-04-08
Making a mnemonic conversion platform independent

Panzerfather:
... to share the code for Winforms/Wpf/Gtk#2+3

Android and WinRT wasn't change because due to VS, but at least for WinRT it could also use the code if it will be changed

iOS, Mac and Mac64 use a special mnemonic code which can't be independent

@cwensley: Is there a known bug in Gtk# why the ToolTips in Source/Eto.Gtk/Forms/ToolBar/*ItemHandler.cs where commented?
At least on Windows with Gtk# 2.12.26 all ToolTips are functional.


cwensley:
Thanks for the pull, however I don't think this belongs in Eto.dll.  To share the code between platforms, you can place the source file in Source/Shared, and then link to that from each of the platform assemblies that require it.

Having this exposed in the Eto namespace will be confusing as it is only used for certain platforms (Gtk and Wpf).

cwensley:
Is there a particular reason to change the partial Conversions class to WpfConversions/MacConversions/etc?  I don't mind changing things around if there's specific reasons behind it, but otherwise it might be good to keep this as is.

Panzerfather:
I think splitting a class over two projects with partial isn't a good Idea because it could confuse people who want to extend Eto and have to search over the project borders to find a method.

But the main reason behind is much more simple than that, the compiler will throw an error that there are to references of Conversion. One in Eto.Conversion and on in Eto.*.Conversion.
So you would have to change your method calls not just simple Conversion.To...() and WinConversion.To...(), you have to call them Eto.Coversion.To...() and Eto.Win.WinConversion.To() which would be annoying.

Only one file SDConversion.cs in the Mac-Solution stays unrenamed and extends the Eto.Conversion-Namespace.
Why? Because it uses the System.Drawing class and this can't be implemented in Eto-PCL directly. And it is spreaded over all other projects which are using the Drawing-Conversion.

cwensley:
I think you're mixing your branches up here, this is now including stuff from pull #321..

cwensley:
Ah, I see you reverted..  would you be able to squash your commits or reset this branch and force push so it doesn't include the unnecessary commits?

Panzerfather:
Just done it!

cwensley:
Hey this is looking good.  Only small thing for me is that NSConversions.cs and CGConversions.cs both contain partial MacConversions - it'd be better to have the name of the files changed to reflect the name of the class to avoid any confusion e.g. MacConversions.ns.cs and MacConversions.cg.cs

Could you also read the [Contributing Guide](https://github.com/picoe/Eto/wiki/Contributing) (especially the part at the bottom about licensing), and remark in a comment that you agree to license the code the same?

Thank you!

Panzerfather:
For NSConversions.cs I agree in renaming, but for CGConversions.cs I disagree, because this file is used/referenced over all platforms (Win/Gtk/...) and extends the Conversions-Namespace.

But this can't be moved to the Eto-PCL code because of the System.Drawing dependency. Making a new Project which only have this file in is a to big overhead I think.

So renaming this to MacConversion.sd.cs isn't possible.

cwensley:
CGConversions is only used by Mac/iOS platforms.. I think you mean SDConversions, which I agree should stay as is.  To remark that you accept it, adding a comment to the pull request is sufficient.

Panzerfather:
@cwensley: I'm accepting the BSD-3 License for this project to this and all other pull requests.

Source/Eto.Mac/MacConversions.cs
```diff
@@ -49,16 +49,10 @@ public static Color ToEto(this NSColor color)
 		{
 			if (color == null)
 				return Colors.Black;
-			var converted = color.UsingColorSpace(NSColorSpace.CalibratedRGB);
```
cwensley: This code shouldn't be reverted here

Panzerfather:
Updated code to your revision c15e0c713ebeb64cd5c7e20d2409bd175a511a55

cwensley:
Great, thanks!!  Though it looks like there's a compile error.   Also, any chance to get a squash of the commits for this one?

Panzerfather:
Puuh, finally rebased and squashed it with the latest picoe/develop tree.

cwensley:
VERY NICE!  Thank you greatly, it looks awesome!

Panzerfather:
No problem. Thank you for your sharp eyes which are watching all the pull requests to keep Eto on a high level programming standard.

PR #329, opened 2015-04-09
Fix for Type "Eto.ResourceNotFoundException" is not marked as serializable

Panzerfather:
Removing EtoException, ResourceNotFoundException and HandlerInvalidException and replacing Exception-Handling with System.Exception to make Eto serializable.

Serialization isn't available in PCL-Code, the used stub serialization isn't functionable. Microsoft supposes using [DataContract] for PCL-Code, but then a whole Error-System have to be created for Eto which is a bit too exorbitant for just printout errors which was using the System.Exception class.

Fixes: Type "Eto.ResourceNotFoundException" in assembly "Eto, Version=2.0.*.*, Culture=neutral, PublicKeyToken=null" is not marked as serializable.


cwensley:
Thanks again for another pull!  

When are you running into the serialization issue?  I'd like to understand what you're doing that requires the exception to be serializable.

I also think if we are going to do this, we shouldn't be throwing Exception, but some other suitable system exception.  See https://msdn.microsoft.com/en-us/library/vstudio/ms229007(v=vs.100).aspx.  For example, we should probably use InvalidOperationException for most of these..

Panzerfather:
This should run you into the exception. It creates an isolated domain and invokes it.

```C#
using System;
using Eto.Forms;

namespace EtoApp.Desktop
{

	public class Program
	{

		[STAThread]
		public static void Main(string[] args)
		{
			AppDomain Domain;
			Facade DomainObject;

			Domain = AppDomain.CreateDomain("Isolate1", null, AppDomain.CurrentDomain.SetupInformation);
			DomainObject = (Facade)Domain.CreateInstanceAndUnwrap(typeof(Facade).Assembly.FullName, typeof(Facade).FullName);

			DomainObject.LoadApplication();

			AppDomain.Unload(Domain);

		}

		public class Facade : MarshalByRefObject
		{

			public void LoadApplication()
			{

				new Application(Eto.Platform.Detect).Run(new MainForm());

			}

		}

	}

}
```

You're right, System.Exception isn't the best thing to do. Eto should use a mix of InvalidOperationException and ArgumentException, ArgumentNullException, and ArgumentOutOfRangeException for Exception-Handling.

And I think we should depend on .NET 4.5 recommendations and not on .NET 4.0 anymore because mono won't support anything less than .NET 4.5 in their upcomming mono 4.0 release (http://http://www.mono-project.com/docs/about-mono/releases/4.0.0/) and dropped all 2.0, 3.5 and 4.0 Frameworks support.
So we should better look at https://msdn.microsoft.com/en-us/library/vstudio/ms229007%28v=vs.110%29.aspx for implementing.

Mono dropped framework support will also affect Eto's .NET 4.0 support which should then be at least marked deprecated.

cwensley:
I agree on following .NET 4.5 guidelines, I merely linked to the other as it was the first hit on google (;

I would like to remove .NET 4.0 support soon, it was only around as long as it has been since it was impossible for people to compile/debug the .NET 4.5 PCL stuff on linux until recently.  At runtime, the .NET 4.5/PCL version runs just fine on mono 3.x in linux.

Panzerfather:
I integrated the exception throwing according to .NET 4.5 guidelines.

cwensley:
Thanks again for this, the contribution is really appreciated!

Issue #331, opened 2015-04-10
TypeConverters ConvertFrom methods in mono 3.x are failing


Panzerfather:
Mono can't load XAML pages which inherits e.g. Size-Informations. Problem is that the used Culture.TextInfo.ListSeparator is different in Linux than in Windows. (Linux=";" - Windows=",").

```
Could not convert object '300, 300' (of type System.String) to {clr-namespace:Eto.Drawing;assembly=Eto}Size: Cannot parse value '300, 300' as size.  Should be in the form of 'width,height' (-2146233088)

StackTrace:
System.Xaml
  at System.Xaml.XamlObjectWriterInternal.GetCorrectlyTypedValue (System.Xaml.XamlMember xm, System.Xaml.XamlType xt, System.Object value) [0x00000] in <filename unknown>:0 
  at System.Xaml.XamlObjectWriterInternal.StoreAppropriatelyTypedValue (System.Object obj, System.Object keyObj) [0x00000] in <filename unknown>:0 
  at System.Xaml.XamlObjectWriterInternal.OnWriteValue (System.Object value) [0x00000] in <filename unknown>:0 
  at System.Xaml.XamlWriterInternalBase.WriteValue (System.Object value) [0x00000] in <filename unknown>:0 
  at System.Xaml.XamlObjectWriter.WriteValue (System.Object value) [0x00000] in <filename unknown>:0 
  at System.Xaml.XamlWriter.WriteNode (System.Xaml.XamlReader reader) [0x00000] in <filename unknown>:0 
  at System.Xaml.XamlServices.Transform (System.Xaml.XamlReader xamlReader, System.Xaml.XamlWriter xamlWriter, Boolean closeWriter) [0x00000] in <filename unknown>:0 
  at System.Xaml.XamlServices.Transform (System.Xaml.XamlReader xamlReader, System.Xaml.XamlWriter xamlWriter) [0x00000] in <filename unknown>:0 
  at Eto.Serialization.Xaml.XamlReader.Load[winMain] (System.IO.Stream stream, EtoApp.winMain instance) [0x00000] in <filename unknown>:0 
  at Eto.Serialization.Xaml.XamlReader.Load[winMain] (EtoApp.winMain instance) [0x00000] in <filename unknown>:0 
  at EtoApp.winMain..ctor () [0x00000] in <filename unknown>:0 
  at EtoApp.Desktop.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0 
```

[...]

PR #332, opened 2015-04-10
Update converters to remove culture specifics. Fix for issue #331

Panzerfather:
Making converters more rock solid and let them parse with XAML-Code even
under linux.

Replacing culture.TextInfo.ListSeparator which differs from platform to platform to the expected comma value.


cwensley:
Hm, looks like the exceptions are mixed in with this pull.  That being said, I'll probably pull both anyway.. However, it looks like for this one the tabs are off in the converters.

Panzerfather:
Corrected the tabs spaces, was my fault. 

´´´
var converter = new Int32Converter();
return new Size(
	(int)converter.ConvertFromString(context, culture, parts[0]),
	(int)converter.ConvertFromString(context, culture, parts[1])
);
´´´
Do we really need this? Won't it be better to use:

´´´
return new Size(
	int.Parse(parts[0]),
	int.Parse(parts[1])
);
´´´

This should be faster and memory saving.

cwensley:
Cool! one last question about these changes.. I noticed you changed plain strings to use string.Format(CultureInfo.CurrentCulture, "string"), is there a reason for that?

Panzerfather:
There are two reasons for doing this:

First:
The .NET framework tries to localizes these strings if you pass the CultureInfo. This makes sure that number formats and language will match in the exception, even if there is no localization for the string available (i.e. it falls back to English).

Second:
I think printing the error (if translation is available in .NET) in the home language of the user is most times better to understand in multi-language applications. The developer will anyway get the StackTrace and more of the exception.

Panzerfather:
Have you looked at the converter code as I mentioned 2 comments above?

I would like to change them for the reasons I mentioned above and it would also be better readable.

e.g. SizeConverters.cs would look like:

```C#
public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
{
	string text = value as string;
	if (text != null)
	{
		string[] parts = text.Split(new string[1] { "," }, StringSplitOptions.RemoveEmptyEntries);
		if (parts.Length != 2)
			throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, "Cannot parse value '{0}' as size.  Should be in the form of 'width,height'", text));

		return new Size(
		    int.Parse(parts[0]),
		    int.Parse(parts[1])
		);

	}
	return base.ConvertFrom(context, culture, value);
}
```

instead of:
```C#
public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
{
	string text = value as string;
	if (text != null)
	{
		string[] parts = text.Split(new string[1] { "," }, StringSplitOptions.RemoveEmptyEntries);
		if (parts.Length != 2)
			throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, "Cannot parse value '{0}' as size.  Should be in the form of 'width,height'", text));

		var converter = new Int32Converter();
		return new Size(
			(int)converter.ConvertFromString(context, culture, parts[0]),
			(int)converter.ConvertFromString(context, culture, parts[1])
		);

	}
	return base.ConvertFrom(context, culture, value);
}
```

cwensley:
string.Format does not actually translate constant strings at all - it only formats numbers, dates, etc when the placeholders are replaced.  So, for all of the constant strings that have no date or number placeholders, having string.Format would be redundant afaik.

For changing the converters, that's no problem - though the idea here is to re-use the converters so that any specific converter logic is reused along the way.  

Removing the culture-specific parts from the converter(s) might actually be the wrong place to do this - instead, perhaps we should change the xaml reader to use an invariant culture?

Panzerfather:
Maybe we misunderstood each other, string.Format with a culture will only printout localized error messages when they are in the .NET runtime, it won't do anything with constant strings if they are not in the runtime only with the placeholders in them.

for own constant strings it have no effect. for them it is more to have an constant error messaging format than of mixing them. On the rewrite I've seen a lot of mixups. Most of them are missing the the CultureInfo when they could use it, some of them using the + sign instead of string.format at all.
So I decided to get all of them the format = string.format(cultureinfo, text).
The performance impact of these is marginal, on an 100000 iteration on a Core i5 VM it's 12-16ms. Memory consumption is the same.

I only spared them out if the error message was: "ugh" or something like this which happens 2-3 times.

The performance impact of the Converter-Change is much more (100000 iteration):
Current implementation: 68-150ms
Suggestion: 28-60ms, it was always at least more than 2.1 times faster and with a memory saving

e.g. Int.Parse can also be Culture specific, but as long as Eto has a fixed format this has only an impact on the error messages not on the conversion of the numbers itself.

You can't run the conversions with invariant culture as supposed because you would loose the value transformation of the running UI-culture (mostly affects values such as DateTime).

firda-cze:
I just don't get the `string.Format(CultureInfo.CurrentCulture, ...)` thing. According to MSDN, providing no `IFormatProvider` is exactly the same as passing `CultureInfo.CurrentCulture`. So, what is the exact difference?

https://msdn.microsoft.com/en-US/library/system.string.format%28v=vs.110%29.aspx

Formatting and culture
----

Generally, objects in the argument list are converted to their string representations by using the conventions of the current culture, which is returned by the CultureInfo.CurrentCulture property.

Panzerfather:
Your right, CultureInfo.CurrentCulture is automatically set when calling string.format().
But this feature is only documented for .NET 4.5 and higher. To ensure that .NET 4.0 will also us CultureInfo.CurrentCulture the string is called with it. This could be removed after Eto will disband .NET 4.0 in the near future, but it has no performance effect at all.

firda-cze:
Well, I cannot find exact documentation for .NET prior 4.5, but am using it this way since 2.0 and it was always implemented (check with ILSpy) with [StringBuilder](https://msdn.microsoft.com/en-US/library/tcz1y0c1%28v=vs.90%29.aspx) while passing `null` which is well documented:

*If the provider parameter is null, format provider information is obtained from the current culture.*

Do you have some experience with it ever doing something different? I am asking just from curiosity, to know, because most of my apps are still targeting .NET 3.5.

cwensley:
Hm, any way we can get a rebase here?  It's hard to review all the exception and conversion changes at the same time..

cwensley:
And we may be able to improve performance and use the converters by keeping an instance of the child converter as a member, instead of instantiating it for every conversion

Panzerfather:
@firda-cze:
For .NET itself I can't remember, but on mono there was a problem around a 2.x.x version (don't remember which exactly) with a 2.0 project so we decided to use this pronunciation. This compatibility will change dramatically in mono 4.0 by using most of Microsoft's .NET Core and the only supported .NET target platform will be 4.5. So keep this in your head before distributions will update to mono 4.0.

@cwensley:
I will make a rebase and a squash later on. And make another performance comparison. But the speedup isn't my main target, it's the memory consumption which was also reduced. If Eto wants to target Android/iOS/... this should be as small as possible.
This memory boost is only if we use .NET's own converters, so I append a try|catch in the latest patch to have even a better reporting functionality than before and removing the cascading approach.

If we remove the Int32Converter design (in my next patch) we can also get rid of the Int32Converter class and the whole Int32ArrayConverter.cs which wasn't used anyway but needed Int32Converter base class. So the memory footage and code of Eto will be even more reduced makes it more readable and reducing the cascading debug if an error occurs.

Panzerfather:
Here are the data of performance how long XamlReader.Load(this) needs to load a really simple Window with 12 Menu-Buttons, 6 Tables [only for layouting], 3 Groupboxes [only for layout], 4 Image-Buttons (average of 10 starts):

With current layout: ~170ms
With new layout (commit): ~145ms

So the boost is impressive and will even become greater in cases where more controls are used which will be the case on 99% of the applications.

Squashed the commits to one.

Source/Eto/Drawing/PaddingConverter.cs
```diff
@@ -12,6 +12,11 @@ namespace Eto.Drawing
 	public class PaddingConverter : TypeConverter
 	{
 		/// <summary>
+		/// The character to split up the string which will be converted
+		/// </summary>
+		private readonly string[] StringSplitter = new string[1] { "," };
```
cwensley: These could be static in each of the classes, and you don't need to use the 'private' modifier as it is redundant.

cwensley:
Looks excellent!  I didn't realize that Int32Converter is no longer used (I believe it was used for old TableLayout api's that are now gone), so thanks for catching that and removing it.

Source/Eto/Drawing/PaddingConverter.cs
Panzerfather: You are right, sadly it had no impact on the performance.

Panzerfather:
No problem, I will look for more unused code in the future.

Source/Eto/Drawing/PaddingConverter.cs
firda-cze: I'd prever `char[]` instead of `string[]` but this is getting..... ugh..... funny polishing :D

cwensley: hm, wonder if there's a perf penalty in specifying a string instead of char

firda-cze: I don't think this really matters. It is just one *test and conditional jump more* which is like nothing for the processor. It could even be `static char Separator = ','` and it would work (while new single-element array would be constructed each time unless they decide to optimize it by adding overload for split-by-sinle-char as they have it for format).

C# is more about code readability, about productivity. You'd have to use `IndexOf` and maybe `unsafe - fixed` to get better performance. Is it worth it?

cwensley: I totally agree about readability and productivity.  The less superfluous code, the better.  

Typically I only sacrifice readability for performance when the impact can be measured.  In this case, these converters can be used thousands of times (depending on your form), so performance is definitely more important and it's nice to see @Panzerfather have actual measurements.

That being said, it would actually be a ton better to have the xaml processing done during compile time so that it has no impact on performance, but that is a problem for another day. (;

Panzerfather: The impact of a string[1] against a char[1] is in such a way marginal that it's hard to measure. To count a time: ~2ms for 1.000.000 iterations.
The reason in .NET is that string is not a type like char or int, it's a referenced type like an object or class. So it's referencing to char with some special implementations (e.g. comparisons). But even memory footprint should be nearly the same (16bit for char) in this case.
However I decided to changed the type to char like @firda-cze supposed.

cwensley: Hm, sorry thought this was more of a academic discussion and pulled before you made the change.

Panzerfather: No problem, I will push this small fix in one of the next pull requests! ;)

Issue #334, opened 2015-04-12
ToolButtonItem in WPF doesn't grayout when disabled


Panzerfather:
The ToolButtonItem isn't grayed out on WPF if it is disabled.

Panzerfather:
Patch are submitted.

PR #335, opened 2015-04-12
Fixes issue #334

Panzerfather:
The grayed out function of a ToolButton doesn't only depends on the
Control.IsEnabled. It also needs the image to be disabled to show it.
Include source code style cleanups.


cwensley:
Awesome, looks good!  Thank you!

PR #337, opened 2015-04-12
Splitter.RelativePosition

### LINKED ISSUE #309: Splitter initial position with FixedPanel=None

firda-cze:
replaces #311 and #310, fixes #309; Copy from #311:

*WinForms should be perfect - I am developing for WinForms for quite some time.
WPF looks good, but I still don't know all the corner-cases and am not sure about the events (will check later and possibly add more unit tests).
GTK2 should work fine, it is mostly based on WinForms implementation. GTK3 has different sizing mechanism, so, I guessed it (cannot curretnly test, but I will probably create some Linux VM in near future). SplitterWidth is guessed as well. I am not so familiar with GTK.*
***There is a stub for Mac, that should not break existing code, but new features are up to you.***
*ThemedSplitter is unfinished, tested under WinForms/D2D, but is not auto-sizing.*

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
@@ -11,37 +11,126 @@ namespace Eto.GtkSharp.Forms.Controls
 		SplitterOrientation orientation;
 		SplitterFixedPanel fixedPanel;
 		int? position;
+		double relative = double.NaN;
+		int suppressSplitterMoved;
+
+		private int _prefer(int width1, int width2)
+		{
+			if (position.HasValue)
+				width1 = position.Value;
+			else
+			{
+				if (!double.IsNaN(relative))
+				{
+					if (fixedPanel == SplitterFixedPanel.Panel1)
+						width1 = (int)Math.Round(relative);
+					else if (fixedPanel == SplitterFixedPanel.Panel2)
+						width2 = (int)Math.Round(relative);
+					else if (relative <= 0.0)
+						width1 = 0;
+					else if (relative >= 1.0)
+						width2 = 0;
+					else
+					{
+						// both get at least the preferred size
+						return (int)Math.Round(Math.Max(
+							width1 / relative, width2 / (1 - relative)));
+					}
+				}
+			}
+			return width1 + width2 + SplitterWidth;
+		}
 
 		class EtoHPaned : Gtk.HPaned
 		{
+			public SplitterHandler Handler { get; set; }
+
 			#if GTK2
 			protected override void OnSizeRequested(ref Gtk.Requisition requisition)
 			{
-				base.OnSizeRequested(ref requisition);
-				if (PositionSet && Child1 != null)
-				{
-					var childreq = Child1.Requisition;
-					if (childreq.Width > 0)
-						requisition.Width += Position - childreq.Width;
-				}
+				var size1 = Child1.SizeRequest();
+				var size2 = Child2.SizeRequest();
+				requisition.Height = Math.Max(size1.Height, size2.Height);
+				requisition.Width = Handler._prefer(size1.Width, size2.Width);
+			}
+			#else
+			protected override void OnGetPreferredWidth(out int minimum_width, out int natural_width)
+			{
+				int min1, width1, min2, width2, sw = Handler.SplitterWidth;
+				Child1.GetPreferredWidth(out min1, out width1);
+				Child2.GetPreferredWidth(out min2, out width2);
+				minimum_width = Handler._prefer(min1, min2);
+				natural_width = Handler._prefer(width1, width2);
+			}
+			protected override void OnGetPreferredWidthForHeight(int height, out int minimum_width, out int natural_width)
+			{
+				int min1, width1, min2, width2, sw = Handler.SplitterWidth;
+				Child1.GetPreferredWidthForHeight(height, out min1, out width1);
+				Child2.GetPreferredWidthForHeight(height, out min2, out width2);
+				minimum_width = Handler._prefer(min1, min2);
+				natural_width = Handler._prefer(width1, width2);
 			}
 			#endif
+
+			protected override void OnSizeAllocated(Gdk.Rectangle allocation)
+			{
+				var it = Handler;
+				if (it == null || double.IsNaN(it.relative))
+				{
+					base.OnSizeAllocated(allocation);
+					return;
+				}
+				it.suppressSplitterMoved++;
+				base.OnSizeAllocated(allocation);
+				it.SetRelative(it.relative);
+				it.suppressSplitterMoved--;
+			}
 		}
 
 		class EtoVPaned : Gtk.VPaned
 		{
+			public SplitterHandler Handler { get; set; }
+
 			#if GTK2
 			protected override void OnSizeRequested(ref Gtk.Requisition requisition)
 			{
-				base.OnSizeRequested(ref requisition);
-				if (PositionSet && Child1 != null)
-				{
-					var childreq = Child1.Requisition;
-					if (childreq.Height > 0)
-						requisition.Height += Position - childreq.Height;
-				}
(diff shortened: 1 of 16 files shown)


cwensley:
This looks awesome!  A few minor things: 

1. can the GTK logic for the HPaned and VPaned sizing be shared somehow?  It appears as if there's the same logic repeated around four times.  Perhaps a static method that both can use to do the calculations?

2. The default.build file should probably be removed (looks like a nant build script?)

firda-cze:
1. I had common method in handler (same as in WinForms) but it was just big `if` according to orientation. This is better (unless you create small methods like `GetAvailableSize` and `GetSizePerOrientation` and such to make the logic same - extracting `Width`/`Height` as needed by orientation could allow one method to handle the logic ... or many ternary operators).
2. Yes, I have used it for backup zipping (before hand-merging code from PositionMode branch). It can be removed. Or it could be enhanced to perform build and testing.... but you have your appveyour script already.

BTW: Why was this pull not checked by AppVeyor?

firda-cze:
About common code: I was thinking about even better solution before (while implementing PositionMode) - share some code in `Splitter` accross platforms (e.g. `double CalcRelative(Size size, double pos, SplitterFixedPanel fix)`). All of the implementations do similar math (some WPF prefers float-points, SWF/GTK ints, but the logic is otherwise same).

Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
```diff
@@ -11,37 +11,193 @@ public class SplitterHandler : GtkContainer<Gtk.Paned, Splitter, Splitter.ICallb
...
+		double relative = double.NaN;
+		int suppressSplitterMoved;
 
 		class EtoHPaned : Gtk.HPaned
 		{
+			public SplitterHandler Handler { get; set; }
+
 			#if GTK2
 			protected override void OnSizeRequested(ref Gtk.Requisition requisition)
 			{
```
firda-cze: Overriding `OnSizeRequested` in each Gtk.Paned removes the big `if (orientation = ...) {} else {}`

Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
```diff
@@ -11,37 +11,193 @@ public class SplitterHandler : GtkContainer<Gtk.Paned, Splitter, Splitter.ICallb
...
+			protected override void OnGetPreferredWidth(out int minimum_width, out int natural_width)
+			{
+				int min1, width1, min2, width2, sw = Handler.SplitterWidth;
+				Child1.GetPreferredWidth(out min1, out width1);
+				Child2.GetPreferredWidth(out min2, out width2);
+				if (Handler.position.HasValue)
+				{
+					var pos = Handler.position.Value;
+					minimum_width = pos + sw + min2;
+					natural_width = pos + sw + width2;
```
firda-cze: GTK#3 calculates both minimal and prefered, which makes it different. Common code would either be called twice (once for minimal and once for preferred) or have the logic even for GTK#2 where it is not needed. GTK#3 does not need the `Math.Max(size1, size2)` (it was left on native) but it is there for GTK#2.

firda-cze:
Nant script removed (last commit amended).

firda-cze:
I have finally managed to run GTK#3 under my new Lubuntu VM and the code seems to have some problems with sizing. Auto-sizing not always as expected (simple test works, but not the complex one with the tree and combo) and the minimal size is quite odd, it really won't allow to schrink the window below "default" and when it is at minimal size and you try to move the splitter/gutter to schrinkit it even more: the window/form gets actually larger! Very strange.

And Eto.Test.csproj needs update according to Eto.Test - pcl.csproj (add splitter unit tests).

I will try to fix that GTK# when I have time and think about some shared code across platforms - in Shared folder.

Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
```diff
@@ -11,37 +11,193 @@ public class SplitterHandler : GtkContainer<Gtk.Paned, Splitter, Splitter.ICallb
...
 				{
-					var childreq = Child1.Requisition;
-					if (childreq.Width > 0)
-						requisition.Width += Position - childreq.Width;
+					var relative = Handler.relative;
+					var fixedPanel = Handler.fixedPanel;
+					if (!double.IsNaN(relative))
+					{
+						if (fixedPanel == SplitterFixedPanel.Panel1)
+							min1 = width1 = (int)Math.Round(relative);
```
firda-cze: Looks like the `min` should not be altered by position and some code to adjust position to honor minimal sizes will be needed.

firda-cze:
What is wrong with the last commit that it cannot be merged?
Can GitHub somehow give me something more descriptive than ugly NO/Unable?

This was the fix for GTK#3 which is reusing some code for both GTK as you requested.
GTK#3 behaves differently when you reach minimal size - GTK#2 will allow to shrink more, GTK#3 will not, until you move the splitter.

Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
```diff
@@ -11,37 +11,126 @@ public class SplitterHandler : GtkContainer<Gtk.Paned, Splitter, Splitter.ICallb
 		SplitterOrientation orientation;
 		SplitterFixedPanel fixedPanel;
 		int? position;
+		double relative = double.NaN;
+		int suppressSplitterMoved;
+
+		private int _prefer(int width1, int width2)
```
firda-cze: This could be nicely reused in WinForms as well, but that would need some common helper, or shared source. What do you think? Struct or class? Class is easier to use, struct has better footprint but using it is a bit harder (some methods need to be static with ref Helper me). I'd prefer class - struct would be overkill.

Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
```diff
@@ -11,37 +11,126 @@ public class SplitterHandler : GtkContainer<Gtk.Paned, Splitter, Splitter.ICallb
...
+		{
+			if (position.HasValue)
+				width1 = position.Value;
```
firda-cze: This could be nicely reused in WinForms as well, but that would need some common helper, or shared source. What do you think? Struct or class? Class is easier to use, struct has better footprint but using it is a bit harder (some methods need to be static with ref Helper me). I'd prefer class - struct would be overkill.

firda-cze:
On second thought, it would need complete redesign of whole framework to reuse code (name same things the same - once it is `UserDesiredSize`, the other time `PreferredSize`) .... I am finished, unless you want something.

firda-cze:
Rebased again and squashed

cwensley:
This is really great, thanks for the squash!  I'll try to fix the build.

I'll try to look at the Mac side of things soon, but I'll add an issue for that so it isn't forgotten.

Issue #339, opened 2015-04-17
ToolBar: Restriction only one toolbar per window


Panzerfather:
The ToolBar implementation in Eto right now is restricted to a single one per window.
So the user multiple on the same window, but many windows programs/tools use more than one and so this is a prerequisite for them to be able to port their code to Eto.

So the ToolBar should be integrated to the Panel control instead of the Window control.

TomQv:
How far on the timeline is this at the moment?
It need this feature for project.

PR #340, opened 2015-04-17
Fix for issue #339: Implement ToolBar to Panel instead of Window

Panzerfather:
This is a fix for #339 and moves ToolBar to Panel to allow multiple ToolBars on one Window. See screenshot in Issue.

It also integrates the latest converter updates (char and VS-layout) [7288cc38d057143ae8270c6f200e0b43eb5454cb].

The Mac-Implementation is a "stub", someone with a Mac have to uncomment the code and test it.

Source/Eto.Mac/Forms/MacPanel.cs (line 121)

``` C#
//TODO: if Mac don't support multiple NSToolbar's we need to get the parent and set the toolbar there.
//      In this case we don't need to care about multiple toolbars, they will be automatically overwritten
//      and the programmer itself have to care about if he wants to support Mac-Platform
//Control.Toolbar = (NSToolbar)toolBar.ControlObject;
```


cwensley:
Hey this is really great, thank you!  

I was thinking instead of moving ToolBar property to Panel, to create a ToolBarView control which you can place anywhere on the form.  This would give you much more flexibility on where the toolbar is placed, instead of just at the top of a panel.  It also makes it *much* easier to implement.

On Mac, this control would not be supported, so you can do something like:

```C#
if (Platform.Supports<ToolBarView>())
{
  var tbv = new ToolBarView { ToolBar = myToolBar };
  ...
}
```

Thoughts?

Panzerfather:
Nice idea, a ToolBarView will make it much easier to layout and should fix the WinRT implementation. Also this would allow a lower (better) implementation for this control if the platform natively supports it.

What exactly is the WinRT-Project about? Is this for Windows 8.x (e.g. Store-Apps/Metro) or is this for pure Windows 8.x RT support which is abandoned?

About a Mac implementation, is their generally no ToolBar control available or only on the window? If there is such a control this can't be done from scratch without knowledge of this platform, so this should then be done by someone else.

cwensley:
The WinRT project is for Windows Store/Metro apps, which in Windows 10 should be usable for desktop, tablet, and phone UI's.  In true Microsoft fashion, the *platform* was named Windows Runtime (or WinRT for short), and then came out with the Windows RT tablet confusing the heck out of everyone..

cwensley:
With Mac you are right in that there is only one toolbar per window.  There's no way to create a toolbar otherwise.  However, I do have an idea that with the ToolBarView, it could be implemented in a way that it uses standard controls when used there.  This way you can get the same functionality on all platforms while still taking advantage of the existing native Window.ToolBar implementation.

Panzerfather:
Thanks for the info about WinRT, in the meantime I installed Windows 8 to get an overview about this platform and to be able to compile these platform.

Should we integrate the ToolBarView as a subset of ToolBars equivalent to the ToolStripContainer or as a single control which only shows the toolbar? For Gtk2|3 we could use a table to show a subset of toolbars.

A similar ToolStripContainer implementation will allow a more flexible integration of multiple ToolBars and the End-user could be allowed to change the position of the ToolBar and even the ToolBars which are shown without heavy coding [equivalent to the Splitter control].

For platforms which are only handle a single ToolBar the kind of implementation makes no difference.

Panzerfather:
@cwensley:
Uploaded a first draft implemented as a ToolStripContainer in WinForms only. Could you please take a look at it if there is something to improve before implementing this in other platforms?

Sorry for the determination to implement this with multiple ToolBars without getting your recall, but you will know why when you'll see the Eto.Test-Example which power this implementation will bring.

Here is a first screenshot (Added one ToolBarView with to ToolBars [Dock: Top, Left]) and a Splitter as Content: [ToolBar-Positions are changeable on runtime via Drag/Drop in all Docking-Positions]

toolbarview

cwensley:
Wow looks great!  However, I think there might be another direction we can go with this to make it much more flexible, by:

1. Make a general DockView that is analogous to what ToolStripContainer does.  This could allow any control to be docked, not just ToolBar's
2. Keep ToolBarView as only a single ToolBar, which can be added to the DockView
3. We will need to keep `Window.ToolBar` pretty much as-is to allow for use of the native OS X toolbar, and to make it simpler to have a toolbar without having to place it on your form manually.

What do you think?

Panzerfather:
I love the idea of making a DockView control but I have to look if this can be easily integrated with platform specific controls.
ToolStripContainer can only have ToolStrip's on the Dock-Positions, not normal controls. DockPanel is only available in WPF. Gtk# doesn't have any controls like this, layout/features have to be done by code.

I'm not sure if we can easily integrate the features [drag/drop feature, multiple toolbars in a single dock] the ToolStripContainer offers, or if we have to code it from scratch.
So maybe we should integrate 3 controls for ToolBarContainerView (for multiple ToolBars), a ToolBarView (for a single ToolBar) and a DockView (multiple controls).

The Window.ToolBar will be back in one of the next commits, this was already on my list but want to do more code cleanups.

cwensley:
The ToolStripContainer can have normal controls using the [ToolStripControlHost](https://msdn.microsoft.com/en-us/library/system.windows.forms.toolstripcontrolhost(v=vs.110).aspx) item, so theoretically it can be used as a generic docking panel (I did a wee test and it seems to work).  

Having a DockView type control is probably the right direction for this, as it would be the preferred end goal.  A ToolBarContainerView would just be a stop gap solution with limited usability in my opinion.

Panzerfather:
You are right, ToolStripControlHost could do the trick for WinForms. I searched for such a extension but didn't found any useful. Every day a new learning day! ;)
But now I played already around with it, so we need to extend some base class(es) to allow difference layout style's [LayoutStyle](https://msdn.microsoft.com/de-de/library/system.windows.forms.toolstriplayoutstyle%28v=vs.110%29.aspx). So there is no need for a special ToolBarContainerView control.

I also need to take a look how to implement the features WinForms have in Wpf to get an overview what else will be needed. Implementing the Drag/Drop feature in Gtk# have to be done by scretch.

@cwensley:
Could you please take a look at to the last build [AppVeyor](https://ci.appveyor.com/project/cwensley/eto/build/2.1.0.255) output.

```
error APPX0108: The certificate specified has expired. For more information about renewing certificates, see http://go.microsoft.com/fwlink/?LinkID=241478. [C:\projects\eto\Source\Eto.Test\Eto.Test.WinRT\Eto.Test.WinRT.csproj]
```

So it seems that the WinRT-Building-Key has expired.

Panzerfather:
Here we go, a first draft of the new DockView and the ToolBarView.

dockview

Three problems which I found so far for the TextBox:
- Ignores at lease width & size properties, but this is not related to the ToolStrip, it also doesn't work in pure Control-Handling
- TextBox is resizing on Text-length. This should be related to the first problem,
- it looses focus when typing, this should be related to the second issue

Panzerfather:
Added ToolBarView to Gtk#2|3.

There seems to be a problem with the Toolbar.ShowArrow - Property:

ShowArrow = true:
toolbarview with showarrow

ShowArrow = false:
toolbarview without showarrow

Eto doesn't resize the control correctly when ShowArrow = true. Maybe you know where to look at?

cwensley:
Hm, not sure why GTK would do that so sorry I can't help there.

For the pull requests, It'd be nice if you could refrain from reorganizing the methods/fields/etc in the classes, it makes it pretty much impossible to review the code (;

Panzerfather:
No problem, I found a solution for the Gtk#-problem, so this will be fixed in next commit.

Sorry about the code reorganizations in the pull request, these comes automatically when a pull request is open in the branch. Eto is hardly fragmented when you have to debug it, so I thought to reorganize it.

Two ideas about to solve this: I can make a list with the files which include code changes or I will manually revert the reorganized files which don't include code changes from this branch.

Panzerfather:
ToolBarView is done for all Projects supporting it.

Wpf:
toolbarview wpf

WinRT (Brand new implementation) (Note: images are missing due unimplemented IWpfImage -> not a ToolBarView error)
toolbarview winrt

So this should be feature complete for now.

For Mac there is only a stub implementation which have to be done by someone who owns one.

cwensley:
That's some really great progress!

Some thoughts on the `DockView` by the way:  Instead of adding a Dock property to the Control class, a `DockViewItem` could contain whatever extra metadata is needed for its elements, such as Dock, Control, Position within the dock, etc.

Also, I think reverting the files that were just reorganized would be ideal.  If possible, reverting the reorganization on modified files would be great too, as no one will be able to tell what is changed from the pull.

Panzerfather:
I will split up this pull requests into at least 3 pull requests to get it a lot smaller and to canalize the discussions around. This will help us to cut down the size and improve the readability.

Moving the files that were just reorganized to another branch and another pull request is possible, but I think moving reorganization changes from files that also includes code changes will be to much work to do (but there are only a few).

I will take a look how to implement a DockViewItem in the DockViewCollection. The DockView will be separated from this pull request to an extra one, it's just another control. ;)

@cwensley:
About the code fragmentation and the depending discussion about the reorganization: Maybe you should narrowing the guidelines about code submissions to get a readable and standard look how the code have to be implemented. This would help a lot to cut down debug time, especially when working with handlers/callback. Right now, it's hard for a third person to read the code and find what you need.

e.g. some examples

This is neither easy readable nor debugable:
```C#
public Control Content { get { return Control != null ? Control.Content : null; } set { Control.Content = value; } }
```

And should look like this:
```C#
public Control Content
{
	get { return Control != null ? Control.Content : null; }
	set { Control.Content = value; }
}
```

For a class I propose a form like this (in example of ToolBarViewHandler.cs):
```C#
// using classes here, ordered by A-Z
using System;
using Eto.Drawing;
using Eto.Forms;
using sd = System.Drawing;
using swf = System.Windows.Forms;

namespace Eto.WinForms.Forms.ToolBar
{
	public class ToolBarViewHandler : WindowsControl<swf.ToolStrip, ToolBarView, ToolBarView.ICallback>, ToolBarView.IHandler
	{
		// All variables here
		// Like: Control content;

		public ToolBarViewHandler()
		{
		}

		protected override void Initialize()
		{
		}

		//
		// Now all the other stuff ordered by A-Z
		//
	}
}
```

Panzerfather:
Code is cleaned form reorganization as good as possible. 2-3 smaller changes left are from VS automatic code layouting.
I'm not quite sure if it's worth spending time in separating the DockView implementation to another branch because it's uses some code fragments from this commit and the rebase of this took me hours to do.

Panzerfather:
I've updated/implemented the Mac ToolBarHandler/ToolBarViewHandler as far as I could. Someone with a Mac environment had to check the code and implement the current stubs in the code to make it work again.

cwensley:
Looking really great!  A few notes:

I don't think it makes sense for both ToolBar and ToolBarView to be a Control.  What would happen when someone tries to add a ToolBar as a child of a panel or table on OS X?  Also, I think that Window.ToolBar should remain as the ToolBar type, and not be using the ToolBarView - again, not all platforms will be able to support ToolBarView, so requiring this here would break the platforms that don't support it.

Thoughts?

Panzerfather:
I choosed the transformation for the ToolBar from Widget to Control because only as a Control you have access in the platform specified handlers to functions like content.GetWindowsHandler(), contentHandler.ContainerControl and others. This the only reason I remember why I had to change it.

The change from Window.ToolBar to Window.ToolBarView is because of the fact that most platforms that supporting ToolBar will/can support ToolBarView either. ToolBarView is like a TableLayout which includes multiple controls like ToolBar(s) and TextBoxes, ...

There are two ways of implementing such a control type in a multi-platform framework:

1. Making it only available for platforms which support all underlaying controls in the parent control.
2. Make it available on all platform which support at least one underlaying control in the parent control.

The developer which is using such a type of control should always have in mind that not all functionality can/is fully supported on every platform (e.g. the restriction to only one ToolBar in OS X).

So I decided to choose the second implementation type and thought the platform handling code has to make sure that only supported controls should be shown.
For the OS X implementation this means that the handler code only should show the first ToolBar inside the ToolBarView-Control and ignore all other controls OR [maybe better approach] merging all included ToolBars into one.

Letting the Window.ToolBar a Window.ToolBar will restrict many other platforms (Win/Linux). So which way to choose?

cwensley:
Hi, sorry for taking so long to respond! 

To get access to the native control, you don't need to use the extension methods.  You can do something like ((ToolbarHandler)toolbar.Handler).Control instead.  Having both ToolBar and ToolBarView as a control seems a little redundant. Furthermore, having the ToolBarView.Content as Control seems odd - what happens if they put non-toolbar controls in there?  Is that supported?  Can you put a ToolBar on a form directly without a ToolBarView? 

The API shouldn't allow you to do things that aren't supported without a way to interrogate the API about its capabilities (though ideally you shouldn't need to).  With your proposed solution, you wouldn't know what is and isn't supported.  Having it very loose where you can put any sort of control in a ToolBarView is not a concise api, and will probably cause developers to do things they shouldn't be doing.  A good post that I keep thinking of while designing Eto.Forms is [Falling into the pit of success](http://blog.codinghorror.com/falling-into-the-pit-of-success/) on Coding Horror, which is a good way to look at making sure Eto.Forms is easy and straightforward to use.

I was thinking something like:

```c#
	public class ToolBarView : Control
	{
		public IList<ToolBar> ToolBars { get;  }
		// ...
	}

	public class ToolBar : Widget
	{
		// ...
	}

	public class Window : Panel
	{
		public IList<ToolBar> ToolBars { get; }

		// for backwards compatibility, possibly mark as obsolete
		public ToolBar ToolBar { get; set; } 
	}
```

This provides a way for a window to have multiple toolbars, plus, allow platforms that support it to implement the ToolBarView that you can place anywhere on the form.  Since the ToolBarView does not contain other Control-based objects, it is not a Panel or Container, it is simply a Control.

This keeps the ToolBar as an abstraction of the toolbar contents, without confusing it with a Control.

Thoughts?

Panzerfather:
No problem at all, being busy too! ;)

> To get access to the native control, you don't need to use the extension methods. You can do something like ((ToolbarHandler)toolbar.Handler).Control instead.

You are right, but I was completely wrong why I had to make the change from a widget to a control. It's because of the Content-Property which only can be a Control.

Our both thoughts about this kind of control are completely different, so I'm trying to explain what I should want to provide with it:

I'm not thinking of it to simply show a ToolBar in this control, but rather be a fully "style-able" container control. ToolBarView was conceived to hold also all other supported controls like TextBoxes, Tables, and so on. It was supposed to build e.g. a ribbon-style ToolBar even on platforms which doesn't natively support it or to just include controls which can't be directly integrated in a ToolBar on all platforms.

A small summary which can be achieved with it can be seen in https://github.com/picoe/Eto/pull/340#issuecomment-94357074 . There is a TextBox and a ToolBar implemented in a ToolBarView.

So should I change it to hold only a ToolBar-Widget in it?

Pros for changing:
- Mac implementation doesn't need to be completely changed. It simply doesn't support the ToolBarView.

Neutral for changing:
- ToolBar-Control can't be placed anywhere on a form, like in upstream (only a problem for Mac)
- Code for Win/Gtk will stay mostly as they are in upstream using Widget again.

Cons for changing:
- A lost of flexibility this control can achieve. [maybe most of the flexibility which will get lost can be achieved when using the new DockView-Control (this has to be ported to other platforms as WPF, but I'm not quite sure that this kind of control can be implemented in WinRT at all)]

> Having both ToolBar and ToolBarView as a control seems a little redundant.

Not really, when you have the new DockView-Control in mind. And of course it will be needed to change a ToolBar-Widget to a full type of control which can be placed anywhere. [+See lines above]

```C#
new DockViewItem
{
	Content = new ToolBarView { Content = this.CreateToolBarTop(), Dock = DockPosition.Top },
	Dock = DockPosition.Top,
	Order = 1,
	Position = new Eto.Drawing.Point(0,0)
},
new DockViewItem
{
	Content = new ToolBarView { Content = this.CreateToolBarRight(), Dock = DockPosition.Bottom },
	Dock = DockPosition.Right
},
new DockViewItem
{
	Content = new TextBox { PlaceholderText ="Search", Text = "Search for" },
	Dock = DockPosition.None,
	Order = 2,
	Position = new Eto.Drawing.Point(200,0)
}
```
PS: Don't get confused when some properties are missing in this pull request code, this example is from my toolbar-dev branch which relies on this pull and extends the controls.

> Furthermore, having the ToolBarView.Content as Control seems odd - what happens if they put non-toolbar controls in there? Is that supported?

Yes it is, that's the intended reason of this control. [See lines above]

> Can you put a ToolBar on a form directly without a ToolBarView?

Yes.

cwensley:
Thanks for your quick response, and sorry for my not so quick ones. (;

I'm not sure I quite follow - what exactly is the difference between ToolBar vs. ToolBarView with a ToolBar as its content? Does a ToolBarView allow multiple toolbars?  What does a ToolBar look like vs. a ToolBarView with a ToolBar inside?  It looks like from the implementation (e.g. gtk) that it will simply reject anything other than the ToolBar as its content..

> Not really, when you have the new DockView-Control in mind. And of course it will be needed to change a ToolBar-Widget to a full type of control which can be placed anywhere. [+See lines above]

What does having both ToolBar and ToolBarView as a control have to do with the DockView?  Why can't you just use the ToolBarView to place a ToolBar anywhere?

I think a more supportable way of allowing any control on a toolbar is to have something like a `ControlToolItem` which would allow you to to specify a Control as its content.  You could then put this on a toolbar on any platform.

Sorry for being so difficult..  I'm trying to wrap my head around the diffs and how to implement this for Mac. This is also a pretty critical feature so I want to make sure we get this right.

Source/Eto.Wpf/Forms/ToolBar/ToolBarViewHandler.cs
```diff
@@ -0,0 +1,120 @@
...
+
+		public virtual Padding Padding
+		{
+			get { return this.Control.Padding.ToEto(); }
+			set { this.Control.Padding = value.ToWpf(); }
+		}
+		
+		public virtual void SetContent(sw.FrameworkElement content)
+		{
+			Control = content as swc.ToolBar;
```
cwensley: Wpf also only supports a ToolBar inside a ToolBarView it appears..

Source/Eto.WinForms/Forms/ToolBar/ToolBarViewHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+		public virtual Padding Padding
+		{
+			get { return this.Control.Padding.ToEto(); }
+			set { this.Control.Padding = value.ToSWF(); }
+		}
+
+		protected virtual void SetContent(swf.Control contentControl)
+		{
+			contentControl.Dock = swf.DockStyle.Fill;
+			this.Control = contentControl as swf.ToolStrip;
```
cwensley: Hm, how does winforms allow you to specify any control as content for ToolBarView?  What if you add a ToolBarView to the form, show the form, then set its content? It appears the ToolBarView instance will get a new Control assigned, but the old toolstrip will still be shown and not updated..

Panzerfather:
> Thanks for your quick response, and sorry for my not so quick ones. (;
As said before, no problem at all. Not a game changing or security issue, just some new features.

> I'm not sure I quite follow - what exactly is the difference between ToolBar vs. ToolBarView with a ToolBar as its content?

ToolBar can only have one ToolBar and only ToolBarItems as content. Just like before.

ToolBarView is thought to have a fully flexible content with ToolBars, TextBoxes, ComboBoxes, usw. in a (possible by user) styled layout. It could be used more like a Ribbon than just a 1:1 ToolBar-Viewer. Therefore it's designed to hold a fully content than just a bunch of ToolBars.

>What does a ToolBar look like vs. a ToolBarView with a ToolBar inside?

The ToolBarView is designed that it can look like whatever you want, wrapping Tables, Controls, etc.around ToolBars. As I pointed out not in this pull because of the code size, but in the next which will follow.

e.g. to let the developer do something like this (source from WPF):
```C#
<ToolBar HorizontalAlignment="Left" Margin="65,119,0,0" VerticalAlignment="Top" Width="406" Height="31">
    <TextBox Height="26" Margin="0" TextWrapping="Wrap" Text="TextBox" VerticalAlignment="Top" Width="120"/>
    <RadioButton Content="RadioButton" VerticalAlignment="Top"/>
    <Grid Height="134" VerticalAlignment="Top" Width="166" Margin="0,0,0,-108">
        <Button Content="Button" HorizontalAlignment="Left" VerticalAlignment="Top" Width="75"/>
        <Button Content="Button" HorizontalAlignment="Left" Margin="81,0,0,0" VerticalAlignment="Top" Width="75"/>
    </Grid>
</ToolBar>
```

>Does a ToolBarView allow multiple toolbars?

Not in this pull, but in one of our offline testing branches. In this pull request ToolBarView is limited to hold only one ToolBar, just as you comment.
But as I said it in my last post, this isn't the end of the line, it's just the beginning of this control.
e.g. in our toolbar-dev branch there are more extensions to the ToolBar and ToolBarView.

All of them haven't been integrated in this pull because of our discussions about the code-size of this pull request.

> What does having both ToolBar and ToolBarView as a control have to do with the DockView?

As I stated out in my last comment and in this, when you think that a ToolBarView isn't just a simple ToolBar-Viewer and we go the way to let ToolBar stay as a widget, we need something adequate to Positioning and Layouting it right without loosing some of the major features.
e.g User repositioning a ToolBar via Drag/Drop on Left/Top/Right/Bottom on a form. The controls implemented right now like a table won't do such a job.

> Why can't you just use the ToolBarView to place a ToolBar anywhere?

This is just a definition of the direction of the venture and what ToolBarView is supposed to be. If it should be only a simple ToolBar-Viewer to be placed on a form, than ToolBarView should be restricted to hold only 1 type of control (ToolBar) and should stay as a widget. Like a pointed out in my last comment.

> I think a more supportable way of allowing any control on a toolbar is to have something like a ControlToolItem which would allow you to to specify a Control as its content. You could then put this on a toolbar on any platform.

Not really or? It was one of my first thoughts before designing the ToolBarView, but with my knowledge right now at least WinForms doesn't support any other type of controls inside a ToolBar when the control is not derived from ToolStripControlHost.
In C# you can't have multiple derivations for one class without using multiple Interfaces. So for every control we would need an Interface for Eto.Forms.Controls and one Interface for ToolStripControlHost. Just something which isn't suit- and handleable. And not a game stopper but something to deal with is the Event-Handling via the OnSubscribeControl/OnUnsubscribeControl-Events. Events won't be exposed/triggered outside the ToolBar without them via these to methods.[https://msdn.microsoft.com/en-us/library/9k5etstz.aspx]
Wpf+WinRT would support it out of the box. I'm not quite sure but I think Gtk# doesn't support all controls either, only Gtk+ does it. Does a Mac-ToolBar support such a control?

> Sorry for being so difficult.. I'm trying to wrap my head around the diffs and how to implement this for Mac. This is also a pretty critical feature so I want to make sure we get this right.

I think we both have the same goal to keep the good code quality Eto provides. So it's good to discuss new features with such huge code changes a little bit longer than others! ;)

We don't run any Mac's so I can't help here, but if Mac will only support a ToolBar in a Widget-Style approach we could go the second option:

- ToolBar stay as a widget
- ToolBarView would only be available on Win+Gtk
- DockView-Control do the design which ToolBarView was supposed to do

Panzerfather:
Maybe you could also take a look at https://github.com/Panzerfather/Eto/tree/ToolBar-dev

In this branch I changed ToolBar to be a widget again and ToolBarView can only hold one ToolBar. Windows can only implement a ToolBar just right now.

So Mac implementation isn't changed and Mac changes in comparison are only layout updates caused from VS. Only one new property (ImageScalingSize) for Mac have to be implemented if supported.

Serg-Norseman:
What is the state of this development now? A very necessary thing!
There is an urgent need to place toolbars in various components, including - vertical toolbars, as well as control the visibility of buttons. Also an important point with the ability to create dropdown buttons with menus, and comboboxes.

cwensley:
Closing as this is way out of date.  #339 tracks this issue going forward.  

One thing to consider, is that we should either make ToolBar itself a Control, *or*, add a ToolBarView.  This PR implements both, which isn't very clear on how to use ToolBar.

We can't (and shouldn't need to) change the API of Window.ToolBar to use ToolBarView (at least for 2.x), it should be ToolBar.  We will also need a full implementation on macOS.

@Panzerfather, just want to note that it is very appreciated that you put the effort into this, the work you did will definitely help in determining how to implement this fully in the future.

Issue #349, opened 2015-06-02
Direct2D Matrix transform produces invalid output


wieslawsoltes:
The following code produces invalid transformations when used in Direct2D platform. The WinForms platform transformations are correct.

    var _gfx = gfx as Graphics;
    ...
    var gp = new GraphicsPath();
    ...
    var m = Matrix.Create();
    var c = new PointF((float)t.CenterX, (float)t.CenterY);
    // translate
    m.Translate((float)t.OffsetX, (float)t.OffsetY);
    // rotate
    m.RotateAt((float)t.RotateAngle, c);
    // skew
    m.Translate(-c.X, -c.Y);
    m.Prepend(
        Matrix.Create(
            1, 
            (float)Math.Tan(Math.PI * t.SkewAngleY / 180.0), 
            (float)Math.Tan(Math.PI * t.SkewAngleX / 180.0), 
            1, 
            0, 
            0));
    m.Translate(c.X, c.Y);
    // scale
    m.ScaleAt((float)t.ScaleX, (float)t.ScaleY, (float)t.CenterX, (float)t.CenterY);

    _gfx.SaveTransform();
    _gfx.MultiplyTransform(m);

    var brush = ToSolidBrush(path.Style.Fill);
    var pen = ToPen(path.Style, _scaleToPage);
    _gfx.FillPath(
        brush,
        gp);
    _gfx.DrawPath(
        pen,
        gp);
    brush.Dispose();
    pen.Dispose();

    _gfx.RestoreTransform();

NuGet package versions (.NET 4.5):
* Eto.Forms, version 2.0.5
* Eto.Platform.Direct2D, version 2.0.5
* Eto.Platform.Windows, version 2.0.5
* SharpDX, version 2.5.0
* SharpDX.Direct2D1, version 2.5.0
* SharpDX.DXGI, version 2.5.0

wieslawsoltes:
Direct2D:
![direct2d](https://cloud.githubusercontent.com/assets/2297442/7937893/c8be4bb4-0942-11e5-9910-441bd83931a5.PNG)

WinForms:
![winforms](https://cloud.githubusercontent.com/assets/2297442/7937900/cf109b8e-0942-11e5-8d2d-2e227dcbc1d1.PNG)

cwensley:
Hm, What values are you using?  I can't reproduce this with 2.0.5 or develop branch.. at least with a simple figure:

[...]

PR #350, opened 2015-06-08
Fix for issue #349: Direct2D Matrix transform produces invalid output

wieslawsoltes:
This is fix for issue #349. The fix works for the PixelOffsetMode.None and the
PixelOffsetMode.Half. 

According to http://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Advanced/PixelOffsetMode.cs,2a8075968e055912
pixels are offset by -.5

### DIFF
--- a/Source/Eto.Direct2D/Drawing/GraphicsHandler.cs
+++ b/Source/Eto.Direct2D/Drawing/GraphicsHandler.cs
@@ -23,8 +23,8 @@ namespace Eto.Direct2D.Drawing
 		bool rectClip;
 		bool disposeControl = true;
 		Bitmap image;
-		float offset = 0.0f;
-		float fillOffset;
+		float offset = 0f;
+		float fillOffset = 0f;
 		sd.Layer helperLayer;
 		sd.Geometry clipGeometry;
 		sd.LayerParameters? clipParams;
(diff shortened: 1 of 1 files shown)


cwensley:
Thanks for the contribution!  Though, I'm not sure this will fix this particular issue.. The reason (from my investigation) that the fill vs. draw is off is because the pixel offset is being applied *after* all other transforms, so the offset is being transformed as well.

I believe the correct fix would be to perform the .5 translation before other transforms..  So in order to change the mode you have to unwind the transform stack, apply the .5 transform, then replay the transform stack.  I believe the same fix will be needed for all platforms except for winforms.

wieslawsoltes:
I think you are right. My fix does not solve the problem completely. The fix solves the offset values being set incorrectly.

cwensley:
First of all, thanks for taking the time to investigate the fix and submitting the pull request.

PixelOffsetMode.None should mean that if I draw a line at (10, 10) to (10, 100), it should appear to be one pixel wide and not be aliased.  PixelOffsetMode.Half would mean to show a single pixel wide line, it would have to be (10.5, 10) to (10.5, 100).

For example, if you run the Eto.Test.WinForms app, the Drawing > Pixel Offset test will show the PixelOffsetMode.None test shows the lines without any blurring effect.  With your change, it shows the lines blurry for the same test.

I will be working soon (next week or so) on a full fix for this issue on all platforms, so I'll let you know how that progresses.

PR #371, opened 2015-08-21
Mac: Fix crash while setting transparent BackgroundColor to MacView

JosefNemec:


cwensley:
Awesome catch.  Could possibly just use the passed in colour, but this is just as effective.  Thanks!

PR #373, opened 2015-08-21
Mac/WPF: Added MenuOpening event to ContextMenu

JosefNemec:


JosefNemec:
Fixed failing build on GTK projects and also added support to WinForms.

cwensley:
Very nice, thanks for the excellent contribution!  

Only small thing is it'd be nice if the MenuOpening event was a handler event, similar to [this](https://github.com/picoe/Eto/blob/develop/Source/Eto/Forms/Controls/Control.cs#L149-L178).  The Handlers would then override `AttachEvent` to attach the event instead of it always being handled even when not needed.

On GTK, you will also have to use a [connector](https://github.com/picoe/Eto/blob/develop/Source/Eto.Gtk/Forms/Controls/PasswordBoxHandler.cs#L17-L45) otherwise the ContextMenu will not be able to be GC'd (yes, I hate doing this too).

Those small things and commenting just the new methods/events in ContextMenu (to get rid of the compile warnings) and it'd be good to go!

Source/Eto.WinForms/Forms/Menu/ContextMenuHandler.cs
```diff
@@ -24,6 +24,20 @@ void HandleOpened(object sender, EventArgs e)
...
+		public override void AttachEvent(string id)
+		{
+			switch (id)
+			{
+				case ContextMenu.MenuOpeningEvent:
+					this.Control.Opening += (sender, e) => Callback.OnMenuOpening(Widget, EventArgs.Empty);
+					break;
+
+				default:
+					AttachEvent(id);
```
cwensley: Missing `base.` here

Source/Eto.Wpf/Forms/Menu/ContextMenuHandler.cs
```diff
@@ -1,14 +1,29 @@
...
+		public override void AttachEvent(string id)
+		{
+			switch (id)
+			{
+				case ContextMenu.MenuOpeningEvent:
+					Control.Opened += (sender, e) => Callback.OnMenuOpening(Widget, EventArgs.Empty);
+					break;
+
+				default:
+					AttachEvent(id);
```
cwensley: base.

cwensley:
Wow, perfect!  Only missing the `base.` in the call to the base class AttachEvent() for each of the platforms, and it'll be done imo.

cwensley:
Wow fast response.  Thanks for helping make Eto better!

PR #375, opened 2015-08-21
Fix for ToolItemCollection items being added in wrong order

JosefNemec:
Happened to me only when toolbar items were being added as Command classes.

### DIFF
--- a/Source/Eto/Forms/ToolBar/ToolItemCollection.cs
+++ b/Source/Eto/Forms/ToolBar/ToolItemCollection.cs
@@ -72,7 +72,8 @@ namespace Eto.Forms
 				else
 					break;
 			}
-			Insert(previousIndex + 1, item);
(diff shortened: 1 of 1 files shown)


cwensley:
Thanks for the pull!  Do you have a repro for the problem you're running into?  This pull request appears like it would break the ability to set the order of the item using the ToolItem.Order property.

cwensley:
Thanks for the pull!  I've committed a fix for this while still keeping the ability to create order groups.  Hopefully this will fix it for you as well.  Let me know if it doesn't. (;

Cheers!
Curtis.

JosefNemec:
Great, sorry that I didn't fixed it my self, I haven't had much time lately.

cwensley:
No problem at all!  Thanks for pointing out the issue. (;

PR #395, opened 2015-10-06
Fallback to Winforms even on Linux and Mac

ermshiperete:
Even though Winforms with Mono is buggy it is still helpful to be able to fallback to Winforms, especially when changing an existing application from Winforms to Eto.Forms.

[Review on Reviewable](https://reviewable.io/reviews/picoe/eto/395)

### DIFF
--- a/Source/Eto/Platform.cs
+++ b/Source/Eto/Platform.cs
@@ -324,8 +324,6 @@ namespace Eto
 				else if (EtoEnvironment.Platform.IsWindows)
 				{
 					detected = Platform.Get(Platforms.Wpf, true);
-					if (detected == null)
-						detected = Platform.Get(Platforms.WinForms, true);
 				}
 
 				if (detected == null && EtoEnvironment.Platform.IsUnix)
@@ -334,7 +332,11 @@ namespace Eto
 					if (detected == null)
 						detected = Platform.Get(Platforms.Gtk2, true);
 				}
-				
+
+				// Fallback to Winforms - better than nothing even if it's not native look & feel
+				if (detected == null)
+					detected = Platform.Get(Platforms.WinForms, true);
+
 				if (detected == null)
 					throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, "Could not detect platform. Are you missing a platform assembly?"));
 					


cwensley:
You can use winforms on any platform by passing it to the `Application` constructor, like so:

```c#
new Application(Eto.Platforms.WinForms).Run(new MainForm())
```

cwensley:
oops, I didn't realize this was a pull request..  Serves me right for looking at it right after getting up and before my coffee. (;

cwensley:
So, I've opened the PR again, but here's the problem I see with falling back to winforms *automatically*:

If it does fallback, people may expect it to work. This is not the case, as I do not test winforms on any other platform than windows.  People have the same confusion when running Eto.Forms apps on Mac from the command line.  They do not realize that it will actually use GTK# unless you bundle it in an .app bundle.  However, at least GTK# apps will run properly on OS X.

For existing applications where you are integrating with existing winforms apps, you would never use the auto detection routines anyway, and just pass the platform you wish to use to the application constructor as I've shown.  You can even perform your own logic to detect which platform to use instead of relying on Eto.Form's auto detection, say for example if you want the ability for the user to specify the platform to use via command line.

ermshiperete:
I wasn't aware that the Application c'tor can take the platform - nice :-) I think it would be helpful if you'd add that to the example given in the readme (change the example to use `new Application(Eto.Platforms.WinForms).Run(new MainForm()`, or at least add a comment) and/or in the tutorial so that people who want to quickly try out Eto.Forms find that information.

> If it does fallback, people may expect it to work. This is not the case, as I do not test winforms on any other platform than windows. 

Fair enough. Having the ability to specify what to use in the c'tor of the Application should be good enough. I'll retract this PR.

Thanks!


---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/picoe/eto/395)

cwensley:
Great, thanks for the feedback and the PR, even if it wasn't merged.  It really helps!  I'll add it to my todo to add information about this to the tutorials and/or readme!   Thanks!!

PR #404, opened 2015-10-20
Fix two OutOfRange exceptions in GridView

ermshiperete:
[Review on Reviewable](https://reviewable.io/reviews/picoe/eto/404)

### DIFF
--- a/Source/Eto.WinForms/Forms/Controls/GridViewHandler.cs
+++ b/Source/Eto.WinForms/Forms/Controls/GridViewHandler.cs
@@ -12,7 +12,7 @@ namespace Eto.WinForms.Forms.Controls
 
 		protected override object GetItemAtRow(int row)
 		{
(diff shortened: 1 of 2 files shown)


cwensley:
Looks great!  Though just wondering in what situations would these exceptions occur?  GetItemAtRow() should never be called with a row < 0 so it'd be nice to know what is causing that.

ermshiperete:
That's a good question and I'm not completely sure of the answer. I was using the GridView in the test app and was wildly clicking around when it crashed. It might have happened after I enabled editing and edited a cell and afterwards clicked on the header row, but I'm not entirely sure and I don't have a Windows machine here to try it out. The immediate cause of the crash was pretty obvious and so I didn't bother checking why we'd get a negative value.

cwensley:
Ok, good to know.  Either way it's good to have these checks to prevent crashes.  Thanks for the contribution!

PR #421, opened 2015-12-15
[XS Addin] Fix missing node on MonoDevelop

cra0zy:
MonoDevelop does not contain app under crossplat, it was removed for whatever reason in the last update. Xamarin Studio contains it.

### DIFF
--- a/Source/Addins/Eto.Addin.XamarinStudio/Properties/Manifest.addin.xml
+++ b/Source/Addins/Eto.Addin.XamarinStudio/Properties/Manifest.addin.xml
@@ -55,6 +55,9 @@
 	<Extension path="/MonoDevelop/Ide/ProjectTemplateWizards">
 		<Class id="Eto.Addin.XamarinStudio.ProjectWizard" class="Eto.Addin.XamarinStudio.ProjectWizard" />
 	</Extension>
+	<Extension path="/MonoDevelop/Ide/ProjectTemplateCategories/crossplat">
+		<Category id="app" name="App" />
+	</Extension>
 	<Extension path="/MonoDevelop/Ide/ProjectTemplateCategories/crossplat/app">
 		<Category id="eto" name="Eto.Forms" />
 	</Extension>


cwensley:
Hi @cra0zy,

Thanks for the contribution!!  

However, I noticed this as well and believe I already fixed it in [this commit](https://github.com/picoe/Eto/commit/8cfb9149628923003e516f977b8c8a32efcae76e#diff-96de181f309e4baa9e8876bf2db139faL49) (though correct me if I did that wrong).  I haven't updated the addin on addins.monodevelop.com yet though as I am finishing up all the other stuff for the 2.2 release which should come out before EOY.

Cheers!
Curtis.

cra0zy:
> However, I noticed this as well and believe I already fixed it in this commit (though correct me if I did that wrong). I haven't updated the addin on addins.monodevelop.com yet though as I am finishing up all the other stuff for the 2.2 release which should come out before EOY.

Oh yea, you are right, sorry.

Do you by any chance have a build server that compiles the addin?

cwensley:
@cra0zy, hm no I don't.. I should add that to the CI build though, it'd be useful for testing and such.  I'll get that going.  

In the meantime, I've built the latest which you can download [here](http://download.picoe.ca/eto.forms/Eto.Addin.XamarinStudio_2.2.0.0-alpha.mpack).  

Hope this helps!
Curtis.

cra0zy:
Thank you very much.

But... my luck sometimes... I have MonoDevelop 5.10 on my system.. addin requers MonoDevelop 5.10.1 :/

cwensley:
@cra0zy, ah right.. I have to build with the minimum version..  Oldest I can get is 5.9.7, so let's try with that (download is [here](http://download.picoe.ca/eto.forms/Eto.Addin.XamarinStudio_2.2.0.0-alpha2.mpack))

cra0zy:
@cwensley Thank you very very much.

cra0zy:
There is still one problem with the MD addin system that I noticed recently, when:
```xml
<Extension path="/MonoDevelop/Ide/ProjectTemplateCategories/crossplat">
    <Category id="app" name="App" />
</Extension>
```
is done it adds a new category always, that means that if there was an app category already, it will remove it, throw all the templates that were in it to "Miscellaneous/General" node, and add a new one.

T

cwensley:
Really?  It doesn't seem to do that with Xamarin Studio on OS X which definitely has the category and many existing templates..  Perhaps it's the order the addins are loaded?  If you remove (restart md) & re-add Eto.Forms addin do the other templates show up?

cwensley:
Hm, though it has hidden some templates under crossplat/library..  now I'm not sure how this should be done to prevent that..

cwensley:
ack, right this is a PR.. Let me create a separate issue #423

PR #427, opened 2015-12-28
Added handling for more keys

xavierzwirtz:


cwensley:
Thanks for the PR!  This is great!

Do you think you'd be be able to take a stab at supporting these keys for other platforms?  If not, let me know and I'll schedule a task to do that.

Also, make sure you set your editor to tabs vs spaces.  Installing [editorconfig](http://editorconfig.org) for VS will do this for you (;

cwensley:
Closing this as it's been 10 months with no updates.  If we add more keys, we'll need to support the other platforms otherwise it won't work as people expect.

PR #429, opened 2015-12-29
Updated Webkit sharp dll file for Gtk 3

cra0zy:
Turns out Gtk 3 webview couldn't work at all with the existing webkit-sharp.dll files, this updates webkit-sharp.dll to version 2.0 (which works with Gtk 3, I think the old version(1.10) is for Gtk 2, I may be wrong on that).


cwensley:
Thanks for the contribution, but it looks like the webkitgtk3-sharp.dll you found has a different API and thus has compile errors (if you click the red X beside the commit)

The original webkit-sharp for gtk3 was from [this fork](https://github.com/shana/webkit-sharp/tree/gtk3), which is updated from the version of webkit-sharp for gtk2.  It just hasn't ever become official unfortunately.

If this new version is more prevalent (it appears to be, but I haven't looked into it yet), then it might be good to update Eto's code to its new api for gtk3 only.

cra0zy:
I checked the Gtk 2 webkit version, it's 1.1

Gtk 3 webkit can be both 1.10 and 2.0

cwensley:
This is really great, thanks for the contribution!  I'll just need some info:

- Where did you get the new dll from?  I need to know where to get updated versions of the assembly when they are available.
- Do you know what distros have packages for this?  I'd like to ensure we have documentation for others on how to set this up properly.  Any details should be added to the wiki.

Thanks!

cra0zy:
> Where did you get the new dll from? I need to know where to get updated versions of the assembly when they are available.

https://github.com/stsundermann/webkitgtk-sharp

> Do you know what distros have packages for this? I'd like to ensure we have documentation for others on how to set this up properly. Any details should be added to the wiki.

None as far as I know...

cwensley:
Thanks again for the PR!  I had to suppress a few obsolete API warnings due to the fact the projects treat warnings as errors (pr #572), but your commit still merged in no problem.  

Awesome that Ubuntu 16.04 now comes with libwebkitgtk3.0 which is compatible with this PR so distributing GTK#3 apps with a WebView is now much easier.

cra0zy:
> Awesome that Ubuntu 16.04 now comes with libwebkitgtk3.0 which is compatible with this PR so distributing GTK#3 apps with a WebView is now much easier.

I no longer use Ubuntu so I didn't even notice but that is definitely nice. 
http://packages.ubuntu.com/xenial/libwebkitgtk3.0-cil

PR #437, opened 2016-01-07
Drag and Drop feature implementation

JosefNemec:
This is a first draft of drag and drop functionality. It's far from finished and it's generally not for merge. I would like to know mainly if this fits into Eto architecture and if this is approach I should be continuing with.

Stuff that currently works:
- drag and drop of standard control (panels, button etc.)
- drag over and drag drop events
- all Windows backends and GTK (I don't have access to Mac at the moment)

Stuff to do:
- drag and drop of individual items from controls like list views, list boxes and others
- drop support of system files (like dragging file from desktop and dropping it into application)
- dragging customization (drag icon for example)

### DIFF
--- a/Source/Eto.Gtk/DragDropConversions.cs
+++ b/Source/Eto.Gtk/DragDropConversions.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using eto = Eto.Forms;
+
+namespace Eto.GtkSharp
+{
+	public static class DragDropConversions
+	{
+		public static Gdk.DragAction ToPlatformDropAction(this eto.DragDropAction dragAction)
+		{
+			Gdk.DragAction action = (Gdk.DragAction)0;
+
+			if ((dragAction & eto.DragDropAction.Copy) != 0)
+			{
+				action |= Gdk.DragAction.Copy;
+			}
+
+			if ((dragAction & eto.DragDropAction.Move) != 0)
+			{
+				action |= Gdk.DragAction.Move;
+			}
+
+			if ((dragAction & eto.DragDropAction.Link) != 0)
+			{
+				action |= Gdk.DragAction.Link;
+			}
+
+			return action;
+		}
+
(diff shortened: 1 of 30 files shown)


cwensley:
Wow, this looks like a really great start!

Funny enough, I've just run into a need for drag/drop in my project as well.

Only change to the api that I would recommend is to instead of automatically calling DoDragDrop, that this should be accessible through Eto api as well so that you can start the drag operation manually.  It won't always be on any mouse down for example, and you might want to start the drag using right click, or in given states of the object, etc.

As for listview/listbox/etc, dragging can be done simply by providing a way to get the item at at a specific mouse position.  Dropping might be more difficult though.  I think we can leave this as a separate pull.

We'll also need a way to get at extra information for the drag/drop (files as you mention, but also other formats like html, bitmap, text, etc). The api should be similar to how the clipboard works, and possibly even unify the api.  Something along the lines of the DataObject in winforms & wpf perhaps.

JosefNemec:
Thanks for the feedback, I'll try to address all your points.

JosefNemec:
Finally managed to get to this. You now call DoDragDrop method to invoke drag operation instead of it being tied to left click. You can also allow/block drop action via DropOver event as in WinForms/WPF.
	
Drop event args now supply additional information, specifically text and Uri metadata (if you drag data from other application or system objects).
	
Note to WinForms: I don't know why, but drag and drop doesn't work properly if Eto.WinForms.Forms.ApplicationHandler.BubbleMouseEvents is enabled. I was unable to find out why, but it looks like first control that invokes drag captures mouse and all other drag actions are using that control.

tomspilman:
It would be really good to see this issue get resolved soon.  We're having to mix in use of XWT to work around not having drag and drop in Eto.

SlowLogicBoy:
Anyone working on this, because it seems to be abandoned?

TomQv:
Any chance to get this done soon?

cwensley:
@JosefNemec Thanks again for this PR, sorry it took so long but I've finally completed the Drag 'n Drop implementation for all platforms.

Issue #447, opened 2016-01-24
Drawable with content (e.g. StackLayout) auto-sizing and drawing.


firda-cze:
I am currently trying to port my attendance app (used in our company) for eto (to enable support for mac) and wanted to mimic `ToolStrip` from WinForms by `Drawable` with simple `LinearGradientBrush`, but:

1. Autosizing does not work (both bars appear to be zero height if I set `height = -1`)
2. Drawing works only with Wpf, it probably gets over-painted in WinForms and GTK by the `StackLayout`

![eto-drawable-content](https://cloud.githubusercontent.com/assets/11376520/12536216/2d6ea2aa-c29f-11e5-9fee-7f6d258ee09c.png)

	def bar drawable
		onPaint, e paintEventArgs override protected
			var g = e.Graphics
			var rc = new rectangle point.empty, size
			using var b = new linearGradientBrush(
				color.fromGrayscale(0.8f),
				color.fromGrayscale(0.6f),
				new pointF(0, 0), new pointF(0, height))
				g.fillRectangle b, rc

	.ctor
		....
		content = new stackLayout
			horizontalContentAlignment horizontalAlignment.stretch
			items
				new bar
					size new size -1,30
					content = new stackLayout
						orientation orientation.horizontal
						verticalContentAlignment verticalAlignment.center
						padding new padding 6,2
						spacing 10
						items
							accountLabel
							accountBox
							todayLabel
							todayBox
				new bar
					size new size -1,30
					content = new stackLayout
						orientation orientation.horizontal
						verticalContentAlignment verticalAlignment.center
						padding new padding 6,2
						spacing 10
						items
							dateBox
							timeBox
							nowButton
				new bar
					size new size -1,50

BTW: This is written in my custom language (I call B#) which may look a bit similar to Boo (which appeared dead, but the author returned recently).

P.S.: Also, the `DateTimePicker` with `Time` mode is useless in WinForms and I cannot get rid of the seconds in WPF (I don't want seconds, hours and minutes only)

firda-cze:
And this happens if I try transparent background color under WinForms:
(see the black boxes instead of labels)
![eto-drawable-swf-transparent](https://cloud.githubusercontent.com/assets/11376520/12537346/0fd04c9e-c2bd-11e5-8999-1efa03d21d8c.png)

Does not change the GTK2 version under Windows (did not test it under Linux).

firda-cze:
[...]

PR #448, opened 2016-01-24
Drawable with content + WinForms Label

firda-cze:
#447 
By the way, I didn't like how `sd.Graphics` were sometimes prematurely disposed
...but the problem with Label was elsewhere - background is already painted by the control => all references to `BackColor` removed from `OnPaint`


cwensley:
Hi @firda-cze, thanks for submitting the issue and the PR!  A few notes:

I've tried doing as you suggest for `GetPreferredSize` before, but it actually causes problems and is much (MUCH) slower in many cases in my testing so I can't recommend we update to doing it that way (even for Drawable unfortunately).

I'm not sure exactly why the Drawable isn't auto sizing correctly in your case, there is not much difference between it and a regular Panel.  It could be simply that the `AutoSize` and `AutoSizeMode` need to be set, and/or to use `swf.Panel`  instead of `swf.Control`.

Also, what is the purpose for overriding `DisplayRectangle`? It doesn't seem related to the auto sizing issue.

I'd also like to see test(s) added to Eto.Test for this issue so we can ensure this is working properly on all platforms.

As for the other updates, they're great but it helps the review if unrelated changes are kept out or have a specific note regarding why, and if the behaviour is intended to change or it is just code cleanup.

Source/Eto.WinForms/WinConversions.cs
```diff
@@ -403,7 +403,7 @@ public static MouseButtons ToEto(this swf.MouseButtons button)
 
 		public static Graphics ToEto(this sd.Graphics g)
 		{
-			return new Graphics(new GraphicsHandler(g));
+			return new Graphics(new GraphicsHandler(g, false));
```
cwensley: When called from DrawableHandler.OnPaint, this will dispose the graphics object which is incorrect.

firda-cze:
Hi, it has been a while :) I have been busy since those splitters ...if you remember.

1. `GetPreferredSize`: I remember I did already propose some modification for Panel, but did not understand the framework into such depth at that time, so rather written it very very carefully, which could slow things. This time I do understand that both `Panel` and `Drawable` are designed as **signle-content**, not for multiple controls (those are to be added to the content, not to panel/drawable directly. That is the reason why this proposed modification should work.

  The speed problem could be hidden in .NET implementation, as it wold be much better to override `internal` implementation (`GetPreferredSizeCore`), because there is some caching logic hidden ... which most often speed things up, but sometimes cause problems on its own (docking vs. autosizing).

2. Why is drawable not autosizing? Because it was derived directly from `swf.Control` and not from `swf.Panel`. The later is a derivate of `ContainerControl` which derives from `ScrollableControl` which handles the logic properly. `Control` does not (even if you set `AutoSize`). Alternative solution would be to split `Drawable` into `SimpleDrawable` and `DrawableContainer` (or the like), where first won't have any content (not a descendant of  eto `Panel` but `Control`) and the later could use WinForms Panel.

3. Why altering `DisplayRectangle`? Because that is what `ScrollableControl` does and the reason why `Panel` works and `Control` does not. `DisplayRectangle` is used in `DefaultLayout` to propose size. Odd, but true, see for your self.

4. Changes related to Graphics: is your comment an agree or do you oppose? The original code would dispose Graphics in OnPaint, where it should not. I was at first searching for the problem of Label black background and found this and thought that `Parent.OnPaint` is actually used for background painting, so, disposing Graphics in OnPaint could screw it up. ....but the real problem was elsewhere - BackColor got set to all zeros (including Alpha chanel ... but Transparent is actually transparent white, not transparent black) and that caused problems in `TextRenderer`. Removed from `OnPaint` as the background (including transparency) is already handled in `OnPaintBackground`.

Source/Eto.WinForms/WinConversions.cs
firda-cze: The original would dispose it in OnPaint, not this modification. `false` = do not dispose, default = true = dispose.

Source/Eto.WinForms/Forms/Controls/LabelHandler.cs
```diff
@@ -239,12 +239,12 @@ protected override void OnPaint(swf.PaintEventArgs e)
...
 						var r = rect;
 						r.Height = position.Size.Height;
-						swf.TextRenderer.DrawText(e.Graphics, position.Text, Font, r, ForeColor, BackColor, textFormat);
+						swf.TextRenderer.DrawText(e.Graphics, position.Text, Font, r, ForeColor, textFormat);
 						rect.Y += r.Height;
 					}
 					return;
 				}
-				swf.TextRenderer.DrawText(e.Graphics, Text, Font, rect, ForeColor, BackColor, textFormat);
+				swf.TextRenderer.DrawText(e.Graphics, Text, Font, rect, ForeColor, textFormat);
```
firda-cze: Transparent background caused problems and background is already handled by control painting (OnPaintBackrdound)

Source/Eto.WinForms/Forms/Controls/DrawableHandler.cs
```diff
@@ -131,11 +152,8 @@ public virtual Graphics CreateGraphics()
 
 		public virtual void Update(Rectangle rect)
 		{
-			using (var g = Control.CreateGraphics ()) {
-				var graphics = new Graphics(new GraphicsHandler(g));
-
-				Callback.OnPaint(Widget, new PaintEventArgs(graphics, rect));
-			}
+			using (var g = CreateGraphics())
+				Callback.OnPaint(Widget, new PaintEventArgs(g, rect));
```
firda-cze: Original was a double-dispose (once in using, later in GraphicsHandler)

Source/Eto.WinForms/Forms/Controls/DrawableHandler.cs
```diff
@@ -29,6 +29,27 @@ public EtoDrawable ()
...
+				{
+					var rect = ClientRectangle;//base.DisplayRectangle;
+					var padding = Padding;
+					rect.X += padding.Left;
+					rect.Y += padding.Top;
+					rect.Width -= padding.Horizontal;
+					rect.Height -= padding.Vertical;
+					return rect;
+				}
+			}
```
firda-cze: this is what `ScrollableControl` does and what `DefaultLayout` requires to properly propose size.

Source/Eto.WinForms/Forms/Controls/DrawableHandler.cs
```diff
@@ -29,6 +29,27 @@ public EtoDrawable ()
...
+
+			public override sd.Size GetPreferredSize(sd.Size proposedSize)
+			{
+				if (Controls.Count > 0)
+					return Controls[0].GetPreferredSize(proposedSize - Padding.Size) + Padding.Size;
+				return base.GetPreferredSize(proposedSize);
+			}
+
```
firda-cze: There is alwasy atmost one control - the content. But `swf.Control` does not handle the logic, `swf.Panel` would.
`return Padding.Size` as the last line could be better.

Source/Eto.Wpf/Forms/Controls/DrawableHandler.cs
```diff
@@ -80,6 +80,25 @@ protected override void OnRender(swm.DrawingContext dc)
...
+				return base.MeasureOverride(constraint);
+			}
+			protected override sw.Size ArrangeOverride(sw.Size arrangeSize)
+			{
+				base.ArrangeOverride(arrangeSize);
+				var content = Handler.content;
+				if (content != null)
+					content.Arrange(new sw.Rect(arrangeSize));
+				return arrangeSize;
+			}
```
firda-cze: corrects autosizing (taken from `sw.Decorator`)

firda-cze:
Added some visual tests ...and added `TableLayout.HorizontalScaled` and rewrote the layout completely, because I was fighting with the `DynamicLayout` too much (too many nestings needed to properly scale the parts .... something like rowspan would help).

firda-cze:
Expected results:
eto-drawable-test-swf

Wpf+Gtk2: The nested black-white gets smaller.
Gtk2: StackLayout does not render stransparent (over-paints the drawable)

Source/Eto.WinForms/Drawing/GraphicsHandler.cs
```diff
@@ -69,12 +69,12 @@ public GraphicsHandler(sd.Graphics graphics, bool shouldDisposeGraphics = true)
...
+		protected override bool DisposeControl
 		{
-			if (!shouldDisposeGraphics)
-				Control = null;
-
-			base.Dispose(disposing);
+			get
+			{
+				return shouldDisposeGraphics;
+			}
```
firda-cze: This is just code cleanup. `DisposeControl` override has exactly this purpose, so why altering `Dispose(bool)` and doing it another way.

firda-cze:
The easiest way would probably be two base `Drawable` on `Panel` or at least the implementation on `swf.Panel`. Now the code involved:

	// System.Windows.Forms.Panel
	internal override Size GetPreferredSizeCore(Size proposedSize)
	{
		Size sz = this.SizeFromClientSize(Size.Empty) + base.Padding.Size;
		return this.LayoutEngine.GetPreferredSize(this, proposedSize - sz) + sz;
	}

This shows how to do it correctly. `SizeFromClientSize` will handle the border (which `Drawable` does not have). Second line will lead us here:

	// System.Windows.Forms.Layout.DefaultLayout
	private static Size LayoutDockedControls(IArrangedElement container, bool measureOnly)
	{
		Rectangle bounds = measureOnly ? Rectangle.Empty : container.DisplayRectangle;
		....

Which explains my `DisplayRectangle` override:

	// System.Windows.Forms.ScrollableControl
	public override Rectangle DisplayRectangle
	{
		get
		{
			Rectangle clientRectangle = base.ClientRectangle;
			if (!this.displayRect.IsEmpty)
			{
				clientRectangle.X = this.displayRect.X;
				clientRectangle.Y = this.displayRect.Y;
				if (this.HScroll)
				{
					clientRectangle.Width = this.displayRect.Width;
				}
				if (this.VScroll)
				{
					clientRectangle.Height = this.displayRect.Height;
				}
			}
			return LayoutUtils.DeflateRect(clientRectangle, base.Padding);
		}
	}

	// System.Windows.Forms.Layout.LayoutUtils
	public static Rectangle DeflateRect(Rectangle rect, Padding padding)
	{
		rect.X += padding.Left;
		rect.Y += padding.Top;
		rect.Width -= padding.Horizontal;
		rect.Height -= padding.Vertical;
		return rect;
	}

firda-cze:
Switched the implementation to `swf.Panel`. Safer.
Wpf works well, it was my fault in the test to specify size. Autosizing works fine now.

I hope you don't mind those new static methods on `TableLayout` :)

Source/Eto.WinForms/WinConversions.cs
cwensley: Right, but we do not want to dispose when the Paint event is called.. e.g. this is incorrect winforms code:

```c#
var ctl = new Control();
ctl.Paint += (sender ,e) => {
  // do painting

  e.Graphics.Dispose();
};
```

with this change, this is effectively what will happen which will probably cause issues.

cwensley: Hm, actually this is my mistake.. It won't dispose the graphics object unless the user code calls dispose as well.  Disregard my comments (;

firda-cze: I hope we now understand each other, but for sure: We don't want to dispose `sd.Graphics` (as this is the responsibility of the creator of the object = caller of WinForms `OnPaint` - done in `WmPaint` called from `WndProc`)

Original `DrawableHandler.OnPaint` would call `ToEto` and actually mark `sd.Graphics` to be disposed together with `GraphicsHandler` and Eto `Graphics`. Exactly what we **don't** want.

My modification passes `false` to the constructor to correct it (and not dispose `sd.Graphics` which you did not create).

cwensley: Hah, wow.  I am fully going to blame this on baby brain from lack of decent sleep.  After many attempts at telling me that you changed it so it WON'T dispose, I _still_ saw it incorrectly as you removed `false` from the line (I swear I'm not dyslexic either).

I see it correctly now.  

This being said, the change you made would only make a difference if you call `e.Graphics.Dispose()` from Eto's `Drawable.Paint` event which you should not ever be doing anyway, so the change is (should be?) moot.

Again, I apologize for misunderstanding the change you made in the first place!

firda-cze: This is **problem of ownership**, you are converting one disposable to another disposable, but the fact you have disposable object at your hand does not automatically mean you have the ownership and should dispose it. Maybe both `ToEto` and `GraphicsHandler` constructor should have non-default bool parametr to let you state it clearly (wether you own the disposable object and want it destroyed or not).

Actually, the very `OnPaint` of the handler could be written with `using(var g = new Graphics(...))` to dispose what you create (**it is disposable after all, it should be disposed by the creator!** what if the object holds some unmanaged resources to provide new features?!), but you do not own the `sd.Graphics` and should therefore **not pass ownership**.

Another question is: will eto Graphics dispose the original when and only when Dispose is called or will it do the same in destructor (called by garbage collector)? Is the "native" graphics treated as managed or non-managed resource? Luckily, `WidgetHandler` treats it as managed (which is correct) and won't dispose it in destructor, luckily, the code was not a disaster as it looked to be :D

...this time, but what next time?
...it is up to you, but I would prefer non-default bool argument to force anobody to think twice, before using the conversion.

cwensley: Hi @firda-cze, thanks for taking the time to think about how this could be done better.  I agree that both `ToEto` and `GraphicsHandler` ctor should have a non-default bool parameter to specify ownership there, which will make things more explicit.

firda-cze:
I have changed the code to reflect what we agreed to about disposable Graphics.
But found interesting/strange problems while trying to add more Drawable tests.
The first complex test of transparency almost killed the WinForms test, Break All ended in `GetPreferrableSize` .... oh, no, the autosizing problem again (I knew it, autosizing vs. docking does not work well in WinForms!) ...but, there is another problem with the transparency and already spoken of Label: (see transparency test - Gray)

eto-transparent-drawable-wpf
eto-transparent-drawable-winforms

And this is what killed WinForms:
(it should have been 4 coloured semi-transparent boxes with smaller non-transparent box of the same color in each .. BTW: something like `preserveAspectRatio` would also be very nice in `TableLayout`)

			var control = new Drawable
			{
				BackgroundColor = Colors.White,
				Padding = 10,
				Content = new TableLayout(
					true,
					TableLayout.HorizontalScaled
					(
						new Drawable
						{
							BackgroundColor = Color.FromArgb(255, 0, 0, 128),
							Content = new TableLayout(
								true,
								null,
								TableLayout.HorizontalScaled(
									null,
									new Drawable
									{
										BackgroundColor = Color.FromArgb(255, 0, 0),
									}
								)
							)
						},
						new Drawable
						{
							BackgroundColor = Color.FromArgb(0, 255, 0, 128),
							Content = new TableLayout(
								true,
								null,
								TableLayout.HorizontalScaled(
									new Drawable
									{
										BackgroundColor = Color.FromArgb(0, 255, 0),
									},
									null
								)
							)
						}
					),
					TableLayout.HorizontalScaled
					(
						new Drawable
						{
							BackgroundColor = Color.FromArgb(0, 0, 255, 128),
							Content = new TableLayout(
								true,
								TableLayout.HorizontalScaled(
									null,
									new Drawable
									{
										BackgroundColor = Color.FromArgb(0, 0, 255),
									}
								),
								null
							)
						},
						new Drawable
						{
							BackgroundColor = Color.FromArgb(0, 0, 0, 128),
							Content = new TableLayout(
								true,
								TableLayout.HorizontalScaled(
									new Drawable
									{
										BackgroundColor = Color.FromArgb(0, 0, 0),
									},
									null
								),
								null
							),
						}
					)
				)
			};

This is how it looks in WPF:
eto-transparent-drawable-wpf-complex

Source/Eto.WinForms/WinConversions.cs
```diff
@@ -401,14 +401,15 @@ public static MouseButtons ToEto(this swf.MouseButtons button)
 			return buttons;
 		}
 
-		public static Graphics ToEto(this sd.Graphics g)
-		{
-			return new Graphics(new GraphicsHandler(g));
-		}
-
-		public static PaintEventArgs ToEto(this swf.PaintEventArgs e)
-		{
```
firda-cze: This conversion is too dangerous

Source/Eto/Forms/Layout/TableLayout.cs
```diff
@@ -158,6 +224,22 @@ public TableLayout(IEnumerable<TableRow> rows)
 			Initialize();
 		}
 
+		/// <summary>
+		/// Initializes a new instance of the <see cref="Eto.Forms.TableLayout"/> class with the specified rows.
+		/// </summary>
+		/// <param name="yscale">Scale all rows</param>
+		/// <param name="rows">Rows to populate the table.</param>
+		public TableLayout(bool yscale, params TableRow[] rows)
+		{
```
firda-cze: very usefull, I hope `yscale` is good name (`scale` could confuse as it does not scale cells)

Source/Eto/Forms/Layout/TableLayout.cs
```diff
@@ -109,6 +109,72 @@ public static TableLayout Horizontal(params TableCell[] cells)
...
+		/// The difference between Horizontal and HorizontalScaled is that this method sets
+		/// ScaleWidth on each cell.
+		/// </remarks>
+		/// <param name="cells">Cells for the row</param>
+		/// <returns>A new single row table layout with the specified cells</returns>
+		public static TableLayout HorizontalScaled(params TableCell[] cells)
+		{
+			foreach (TableCell cell in cells)
+				if(cell != null)
+					cell.ScaleWidth = true;
```
firda-cze: very usefull and that `if(cell != null)` is required (added a bit later)

Source/Eto.Test/Eto.Test/Sections/Controls/DrawableSection.cs
```diff
@@ -95,11 +125,164 @@ Control LargeCanvas()
...
+
+			return control;
+		}
+
+		Control Transparent()
+		{
+			/*
+			THIS WOULD REALLY KILL THE APP IN WINFORMS (WPF IS GOOD)
+			AUTOSIZING IS HURTING BADLY
+
```
firda-cze: I would really like to keep this comment until it is solved

Source/Eto.WinForms/Forms/Cells/CustomCellHandler.cs
```diff
@@ -98,7 +98,7 @@ protected override void Paint(sd.Graphics graphics, sd.Rectangle clipBounds, sd.
 				if (!ReferenceEquals(cachedGraphicsKey, graphics) || cachedGraphics == null)
 				{
 					cachedGraphicsKey = graphics;
-					cachedGraphics = new Graphics(new GraphicsHandler(graphics, shouldDisposeGraphics: false));
+					cachedGraphics = new Graphics(new GraphicsHandler(graphics, dispose: false));
```
firda-cze: refactoring, now we could use `graphics.ToEto(false)`

firda-cze:
I have to ask, to get the understanding of the framework:
Whenever I try something, it usually works in WPF, but when I try the same in WinForms, it may not work... (well, some things like those drawable did not work in any of the two).. and what does not work is often somehow related to either **transparency or layout or both**. (And GTK seems to work exactly when WinForms work except for imperfect Windows implementation which lacks the transparency completely ... and that is not fault of Eto.Forms but GTK#).

1. Are Labels and Layout intended to be transparent or not? (They are in WPF, perfectly.) Is it really purposedly platform specific and I should always specify transparent background? Even when it defeats nice conversion like string to Label (and to TableCell or Row)? Would transparent Label in WinForms hurt performance?

			public EtoLabel()
			{
				Wrap = WrapMode.Word;
				BackColor = sd.Color.Transparent;
			}

  this solved the problem with Label on semi-transparent Drawable on another Drawable.

2. Are WinForms treated as secondary and possibly not-working platform with the entusiasm towards WPF as leading platform? This could answer or be answered by the previous question.

I have always liked WinForms, it just works... well, except autosizing, but that was usually what I did not need in depth, just autosize the `Label` and `TextBox.Height` and that is all I need, I will handle the rest.... but with so nice cross-platform toolkit like Eto really is *(and I really admire you for doing such a thing)*, are those platform-specific differences intended or have other reasons/purposes like speed? I just don't get it. Why writing such nice and universal toolkit and then force the users to some patterns to actually achieve platform independence?

I hope I was not rude this time, I really really like the toolkit and support your effort, but as much as I like it, that much I want it to work ;) BTW: I am super-Libra (astrology, many aspects in Libra) and I have strong sense for beauty and perfectness.

cwensley:
@firda-cze, thanks for the questions, hopefully I can provide some insight!

1. Yes, I think having control backgrounds as transparent by default is ideal.  There haven't been any tests added regarding what you've been trying to accomplish (using controls as children of Drawable), hence some of the bugs.  I don't think a transparent label would hurt performance on winforms, but it would require testing.

  Just to note, GTK2 lacks control transparency on all platforms (it is just not supported).  Eto.Forms does some trickery on GTK2 to set the background colour of child controls when you set the container's control to 'simulate' transparency, but it obviously is just a stopgap. You'd have to use GTK3 to get true transparency, but I haven't tested it fully.

2. All desktop platforms in my mind have the same priority (even though I have contracts to support WPF and Mac work specifically).  There will still be differences between platforms that are usually not intended, and there will still be some bugs, and some that are not known/recorded yet, obviously.  If there weren't any bugs, I wouldn't have anything to work on so we have to tackle them one at a time.

  I do eventually see a future where the winforms platform is deprecated. With the advent of high DPI displays, WPF just handles all of it much better.  The only thing that winforms has over WPF right now is its immediate mode graphics, which I hope to eventually support under WPF using Direct2D.  It all comes down to what the people want vs. the maintenance overhead.  You are not the only one that leans towards using WinForms, so it isn't going anywhere soon.

One of my concerns that pretty much go along with any PR is testing on **all** platforms.  If a feature only works on one platform (barring [unsupportable functionality](https://github.com/picoe/Eto/wiki/Feature-Detection)), then it renders this library utterly useless.  Keeping all *supported* functionality working across all platforms takes highest priority.

I primarily use the Eto.Test application to determine any platform differences, so only the tests there are things I can safely say work well across all (desktop) platforms.

Hope this helps!
Curtis.

firda-cze:
Thanks for your reply, I will try to continue my contribution within the lines you stated.

Transparent Label and Layout should be good, but I was, in the meantime, thinking about performance and best implementation not to hurt what is present (and believe me, I know how and why to be cautious). So, my assumption/axiom was that only drawable is able to paint non-solid picture/background. So, not to hurt existing code and performance (as drawing backround through endless chain of parents could hurt), I have focused to provide improvements specifically for Drawable (as any other control would already be handled by automatic `BackColor` sharing from `Parent` and its parent and its parent... and so on).

Dispite your warning and experience, I have altered the layout in such a way, that the killing test (described previously - 4 semi-transparent boxes with one non-transparent box in each) showed within few second (still not perfect, but few seconds compared to 25s is huge improvement). **But, I came into design problem: how should `Panel` recognise that it should scroll the content and not propose the size of its content? This is what happened:**

eto-drawable-swf-semi-fix

There still seems to be something wrong with `TableLayout` (and `null`), but I would like to first understand why the `Large Canvas` was not auto-sized to scroll within available size but proposed to be fully visible (**and scrolled by the outer control, not within the `Panel`**).

About other things and cross-platform-ness: I am happy to hear what you stated. I like WinForms, but am willing to switch to default Windows Eto, which is WPF. I require to have default working implementation for each platform: Windows, Linux and Mac. I don't care how, but those three have to have some and default working implementation ;)

firda-cze:
I think I have finally solved the problem:
A: Added `BackgroundColorSet` and altered `SetContent` that I can do:

		protected override void SetContent(Control control, swf.Control contentControl)
		{
			var handler = control.Handler as IWindowsControl;
			if (handler != null && !handler.BackgroundColorSet)
				contentControl.BackColor = sd.Color.Transparent;
			base.SetContent(control, contentControl);
		}

this should be perfectly safe (not hurting performance by using transparent background when not really necessary)

B: Altered the transparent test by specifying size for those boxes

									new Drawable
									{
										BackgroundColor = Color.FromArgb(255, 0, 0),
										Size = new Size(10, 10)
									}

now even WinForms are able to do it in time (although it is not so fast, but acceptable).

eto-transparent-drawable-swf-fix

C: Still, I would like to understand the layout (especially how should the panel decide whether to propose the size of its content or accept anything and scroll the content - in this case that large canvas). Something still stinks about the layout logic in WinForms and I would like to find it. For now, I have ceased my attempts to fix it (and just added those size specifications to the test, to make WinForms pass it).

D: Ammended last commit not to mess things with experiments.

Source/Eto.WinForms/Forms/Controls/DrawableHandler.cs
```diff
@@ -7,26 +7,25 @@
 
 namespace Eto.WinForms.Forms.Controls
 {
-	public class DrawableHandler : WindowsPanel<DrawableHandler.EtoDrawable, Drawable, Drawable.ICallback>, Drawable.IHandler
+	public class DrawableHandler : WindowsPanel<DrawableHandler.EtoPanel, Drawable, Drawable.ICallback>, Drawable.IHandler
 	{
 		public virtual bool SupportsCreateGraphics { get { return true; } }
 
-		public class EtoDrawable : swf.Control
+		public class EtoPanel : EtoPanel<DrawableHandler>
```
firda-cze: Moved common code to WindowsControl.cs not to copy it all the time

Source/Eto.WinForms/Forms/Controls/PanelHandler.cs
```diff
@@ -7,49 +7,13 @@ namespace Eto.WinForms.Forms.Controls
 {
 	public class PanelHandler : WindowsPanel<PanelHandler.EtoPanel, Panel, Panel.ICallback>, Panel.IHandler
 	{
-		public class EtoPanel : swf.Panel
+		public class EtoPanel: EtoPanel<PanelHandler>
```
firda-cze: Common code moved to WindowsControl.cs

Source/Eto.WinForms/Forms/Printing/PrintDocumentHandler.cs
```diff
@@ -49,12 +49,13 @@ public override void AttachEvent(string id)
 
 		protected virtual void HandlePrintPage(object sender, sdp.PrintPageEventArgs e)
 		{
-			var graphics = new Graphics(new GraphicsHandler(e.Graphics));
-
-			var args = new PrintPageEventArgs(graphics, e.PageBounds.Size.ToEto(), currentPage);
-			Callback.OnPrintPage(Widget, args);
-			currentPage++;
-			e.HasMorePages = currentPage < PageCount;
+			using (var graphics = e.Graphics.ToEto(false))
```
firda-cze: this is how it should be done - dispose the disposable if you create it (but don't hurt native graphics by passing `false` to `ToEto`)

Source/Eto.WinForms/Forms/WindowsControl.cs
```diff
@@ -91,6 +93,49 @@ public abstract class WindowsControl<TControl, TWidget, TCallback> : WidgetHandl
...
+				MinimumSize = sd.Size.Empty;
+				AutoSize = true;
+				AutoSizeMode = swf.AutoSizeMode.GrowAndShrink;
+			}
+
+			public override sd.Size GetPreferredSize(sd.Size proposedSize)
+			{
+				var userSize = Handler.UserDesiredSize;
+				var size = userSize.Width >= 0 && userSize.Height >= 0 ? sd.Size.Empty
+					: base.GetPreferredSize(proposedSize);
```
firda-cze: performance improvement, no need to call `base.GetPreferredSize` when we won't use the result

Source/Eto.WinForms/Forms/WindowsControl.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Eto.WinForms.Forms
 {
-	public interface IWindowsControl
+	public interface IWindowsControl: Control.IHandler
```
firda-cze: Added it that we can access `BackgroundColor` as well (all the controls implement `Control.IHandler` anyway)

firda-cze:
Had to disable the feature for Direct2D, drawable in D2D does not support transparent background. I will look into it later, but for now, this commit will fix the exception preventing to even display drawable tests.

eto-transparent-drawable-d2d

Source/Eto.WinForms/Forms/Controls/DrawableHandler.cs
```diff
@@ -11,22 +11,21 @@ public class DrawableHandler : WindowsPanel<DrawableHandler.EtoDrawable, Drawabl
 	{
 		public virtual bool SupportsCreateGraphics { get { return true; } }
 
-		public class EtoDrawable : swf.Control
+		public class EtoDrawable : EtoPanel<DrawableHandler>
```
firda-cze: Moved common code to WindowsControl.cs not to copy it all the time

firda-cze:
It is simply not possible to enable this feature for true D2D,
but I think it is fair to fall back to good old painting system, when content on drawable is set.

For this, I have to rewrite the whole drawing (brushes, pens, ...) to enable both modes.
**Is this good way?**

firda-cze:
At least got the test working, but it still needs work

eto-d2d

firda-cze:
Made the changes `#if !WINRT` to make it build again...
...but, during tests, I often see this (whenever I try something I did not reimplement yet)

    Exception thrown: 'System.InvalidCastException' in Eto.WinForms.dll
    The program '[7456] Eto.Test.Direct2D.vshost.exe' has exited with code -1073741819 (0xc0000005) 'Access violation'.

All the time ... and almost always in `WinConversions.cs` which assume WinForms/System.Drawing objects (pens, brushes, paths). Luckily, Visual Studio is able to at least show the line, before it all crashes witht that **access violation**.

Source/Eto/Drawing/Colors.cs
```diff
@@ -292,8 +292,19 @@ public static class Colors
...
 		/// <summary>Gets a color with an ARGB value of #FFDCDCDC</summary>
 		public static Color Gainsboro { get { return Color.FromRgb(0xDCDCDC); } } // 220 220 220
+
+		// difference between Gray and Grey is very confusing
+		// some beleive that they are equal (while Gray being American and Grey British English)
+		// but while both may mean #808080, Grey sometimes refer to #D3D3D3 or "silver" (#C0C0C0)
+		// see http://www.greyorgray.com/ and System.Drawing.LightGray (no Grey or LightGrey)
+		// Grey should therfore not be present for not being exact and LightGray and LightGrey being aliases
+		// ...although LightGrey is actually wrong, but we keep it for compatibility, sanity and misspelling
+
```
firda-cze: I will probably remove this comment. Although `LightGrey` was the only occurence of `Grey

Source/Eto/Forms/Controls/Drawable.cs
```diff
@@ -197,7 +197,7 @@ public new interface IHandler : Panel.IHandler
 			bool SupportsCreateGraphics { get; }
 
 			/// <summary>
-			/// Creates a new drawable control without specifying a large canvas or not
+			/// Creates a new drawable control without specifying canvas size
```
firda-cze: Another stupid modification to be removed.
But to be honest, maybe there should be some hint as to or not to use D2D on such platforms that support direct drawing (or whatever similar system). Maybe not all drawables are to be painted by D2D (in my PR it will be distinquished by having and not having content).

cwensley:
@firda-cze, thanks for looking into D2D in detail.  I don't think we need to do a fallback to GDI at this point for it, it seems like it might be problematic when creating pens/brushes/paths/etc outside of the paint event.  

Adding a `Drawable.SupportsContent` property would probably suffice for now, and one can select the appropriate rendering system themselves.

firda-cze:
Well, not that much, it is doable. I have already solved most of it
- `SolidBrush`
- `LinearGradientBrush`
- `Pen`
and was going for `GraphicsPath` (was returning empty path in last commit, not to kill app and allow other tests)

All it needs is to provide proper interface, because the concrete `GraphicsHandler` will ask for the proper version to be created / maintined for it (it is just pure data, until GDI or D2D ask for it to do the actual painting).

But it is a lot of work and probably unnecessary... well, I will move it to separate branch (for possible later work) and clean and squash this, that we can move on with th PR.

`Drawable.SupportsContent` would allow to just add `Panel.Handler` I suppose, to some custom universal *decorating* control, to remove the painting and let it at least work.

firda-cze:
Altering the painting flags for transparent background (see TableLayout and WindowsControl.EtoPanel) made HUGE change for drawable tests when you resize the form.

Flickering is gone and even the layout feels like faster.

I can look into more optimizations for WinForms (e.g. for TreeView, it flickers a lot when sizing)... to do so, I have two questions:
1. Did I really cover all layout? How is StackLayout implemented?
2. Are WinForms to be running on Windows only (thus allowing me some hardcore tweaks) or do you plan to support Mono WinForms (Linux, Mac)? ...I can detect the true platform and enable some tweaks for Windows only.

Here is the core of the added magic:

            // optimization especially for content on drawable
            protected override void OnBackColorChanged( EventArgs e )
            {
                SetStyle
                    ( swf.ControlStyles.AllPaintingInWmPaint
                    | swf.ControlStyles.DoubleBuffer
                    , BackColor.A != 255 );
                base.OnBackColorChanged( e );
            }
            protected override void OnParentBackColorChanged( EventArgs e )
            {
                SetStyle
                    ( swf.ControlStyles.AllPaintingInWmPaint
                    | swf.ControlStyles.DoubleBuffer
                    , BackColor.A != 255 );
                base.OnParentBackColorChanged( e );
            }

cwensley:
The updates look great!  Thanks for looking into that.  This will make the winforms platform much more polished. Here's my answers to your questions:

1. The StackLayout uses the TableLayout in the background (for now), so there's only TableLayout and PixelLayout that you have to worry about.
2. WinForms is only tested to run on windows, but I do want to ensure that anything windows-specific (calling into win32, for example), is only run when on windows.  `EtoEnvironment.Platform.IsWindows` can be used for that.

BTW, It looks like this PR is getting a little monolithic and harder to review & test independently.. it is usually a good idea to split PR's to separate unrelated changes (e.g. TableLayout enhancement, Transparent drawable content, double buffering enhancements, etc).   If you have anything else unrelated to these changes you'd like to add to Eto, a separate PR would be very much appreciated.  I do know it'll be a bit more hassle on your end, but I'll be able to merge them quicker as they will be more self contained (;

Again, thanks for all your hard work on this, it'll make Eto.Forms much better for everyone!

firda-cze:
1. Thanks, it seems I have covered it all (and yes, realized that StackLayout is implemented by TableLayout, just wanted you to check it, to be sure)
2. I can do it this way, but you already have `Win32` there and it is not used this way. So, I did it the same way. I have actually only added few constants and called `SendMessage` that was already present, no new P/Invoke.

And now about the BTW: I understand that it is not easy to review bigger work, but it has to be this way. Our time is precious, both yours and mine. I respect your time, you respect mine. Last time (I am refering to those Splitters, it is almost a year back), you wanted me to squash it. So, I did it this time automatically, after we made it clear, what is intended and needed.

I could separate WinForms from Wpf, but I needed those new tests for both. So, imagine it, how am I supposed to actually split it to parts, when it is sequential work, one step dependent on the previous? I could separate the last commit (grid/tree optimization), yes, I could, but that is a drop in an ocean, compared to the monolith, that in principle has to stay big monolith.

Take your time, request more tests, I will do them as this is what I see as good way. I had to implement many automatic tests for my own work (firmware programmer, single universal firmware for endless combinations of hardware, it just became humanly untestable). I will implement whatever test you request to proove this PR is worth accepting, but please, don't ask me to separate it, it's hardly doable.

Source/Eto.Direct2D/Platform.cs
```diff
@@ -13,6 +13,10 @@ public override string ID
 			get { return Platforms.Direct2D; }
 		}
 
+		public override PlatformFeatureFlags SupportedFeatures =>
+			base.SupportedFeatures & ~ PlatformFeatureFlags.DrawableWithTransparentContent;
+
+
```
firda-cze: Not supported on D2D as agreed

Source/Eto.Test/Eto.Test/MainForm.cs
```diff
@@ -32,7 +32,15 @@ public TextArea EventLog
...
-			Title = string.Format("Test Application [{0}]", Platform.ID);
+			Title = string.Format("Test Application [{0}, {1} {2}, {3}]",
+				Platform.ID,
+				EtoEnvironment.Is64BitProcess ? "64bit" : "32bit",
+				EtoEnvironment.Platform.IsMono ? "Mono" : ".NET",
+				EtoEnvironment.Platform.IsWindows ? EtoEnvironment.Platform.IsWinRT
+				? "WinRT" : "Windows" : EtoEnvironment.Platform.IsMac
+				? "Mac" : EtoEnvironment.Platform.IsLinux
+				? "Linux" : EtoEnvironment.Platform.IsUnix
+				? "Unix" : "Unknown");
```
firda-cze: Just more info, but very good, wanted to know this when testing

Source/Eto.Test/Eto.Test/Sections/Controls/DrawableSection.cs
```diff
@@ -9,23 +9,58 @@ public class DrawableSection : Scrollable
...
+					)),
+
+					(Platform.SupportedFeatures & PlatformFeatureFlags.DrawableWithTransparentContent) == 0 ?
+					new TableRow(
+						"(Transparent content on drawable not supported on this platform)"
+					) : null,
+
+					null
+				}
+			};
```
firda-cze: I really needed this to test it properly and thus added `TableLayout.Horizontal[Scaled]` and few more.

Source/Eto.Test/Eto.Test/Sections/Controls/DrawableSection.cs
```diff
@@ -95,11 +130,162 @@ Control LargeCanvas()
...
+			return control;
+		}
+
+		void LogEvents(Drawable control, string name)
 		{
 			control.Paint += delegate(object sender, PaintEventArgs pe)
 			{
-				Log.Write(control, "Paint, ClipRectangle: {0}", pe.ClipRectangle);
+				Log.Write(name, "Paint, ClipRectangle: {0}", pe.ClipRectangle);
```
firda-cze: The log now contains name of the drawable, to separate the problems when more events are fired then expected.

Source/Eto.Test/Eto.Test/Sections/Drawing/BrushSection.cs
```diff
@@ -312,6 +315,22 @@ Control UseBackgroundColorControl()
...
+				TableLayout.AutoSized(
+					new Label() {
+						Text = "Some text",
+						BackgroundColor = Colors.Transparent },
+					centered: true
+				).With(tl => tl.BackgroundColor = Color.FromArgb(0x40000000)) : null;
+			};
+			return control;
+		}
+
```
firda-cze: This tests the transparency of content over drawable vs. brushes.
It can help to implement and test it on other platforms (e.g. D2D).

Source/Eto.WinForms/Drawing/GraphicsHandler.cs
```diff
@@ -62,19 +62,24 @@ public GraphicsHandler()
 
 		bool shouldDisposeGraphics = true;
 
-		public GraphicsHandler(sd.Graphics graphics, bool shouldDisposeGraphics = true)
+		/// <summary>
+		/// Create handler for <see cref="Graphics.IHandler"/>.
+		/// </summary>
+		/// <param name="graphics">Native graphics to use</param>
+		/// <param name="dispose">Pass ownership and dispose native graphics in Dispos></param>
+		public GraphicsHandler(sd.Graphics graphics, bool dispose)
```
firda-cze: Refactoring and description, because converting disposable to another disposable is dangerous. We agreed to state ownership explicitly (non-default `bool dispose`).

Source/Eto.WinForms/Forms/Controls/DrawableHandler.cs
```diff
@@ -11,22 +11,23 @@ public class DrawableHandler : WindowsPanel<DrawableHandler.EtoDrawable, Drawabl
 	{
 		public virtual bool SupportsCreateGraphics { get { return true; } }
 
-		public class EtoDrawable : swf.Control
+		public class EtoDrawable : PanelBase<DrawableHandler>
```
firda-cze: I see no way to review changes to this file just by looking at them. It has to be tested well... and it was, deeply ;)

Source/Eto.WinForms/Forms/Controls/DropDownHandler.cs
```diff
@@ -158,18 +158,6 @@ public DropDownHandler()
...
-			get
-			{
-				return Control.BackColor.ToEto();
-			}
-			set
-			{
-				Control.BackColor = value.ToSD();
-			}
-		}
-
```
firda-cze: This was the same as original base implementation. Removed to enable new transparent background on drawable (as `base.BackgroundColor` now manages `BackgroundColorSet`)

Source/Eto.WinForms/Forms/Controls/TreeViewHandler.cs
```diff
@@ -15,12 +15,30 @@ public class TreeViewHandler : WindowsControl<swf.TreeView, TreeView, TreeView.I
...
+					| swf.ControlStyles.AllPaintingInWmPaint
+					, true);
+			}
+
+			const int TVS_EX_DOUBLEBUFFER = 0x0004;
+			protected override void OnHandleCreated(EventArgs e)
+			{
+				base.OnHandleCreated(e);
+				Win32.SendMessage(Handle, Win32.WM.TVM_SETEXTENDEDSTYLE,
+					(IntPtr)TVS_EX_DOUBLEBUFFER, (IntPtr)TVS_EX_DOUBLEBUFFER);
```
firda-cze: Mono Windows.Forms would never work without adding `SendMessage`, this is acceptable P/Invoke (and I did not add any, just used what was there already)

Source/Eto.WinForms/Forms/TableLayoutHandler.cs
```diff
@@ -58,6 +58,24 @@ protected override void SetBoundsCore(int x, int y, int width, int height, swf.B
...
+				base.OnBackColorChanged( e );
+			}
+			protected override void OnParentBackColorChanged( EventArgs e )
+			{
+				SetStyle
+					( swf.ControlStyles.AllPaintingInWmPaint
+					| swf.ControlStyles.DoubleBuffer
+					, BackColor.A != 255 );
+				base.OnParentBackColorChanged( e );
+			}
```
firda-cze: See WindowsControl.cs

Source/Eto.WinForms/Forms/WindowsControl.cs
```diff
@@ -91,6 +93,79 @@ public abstract class WindowsControl<TControl, TWidget, TCallback> : WidgetHandl
...
+			}
+			protected override void OnParentBackColorChanged( EventArgs e )
+			{
+				SetStyle
+					( swf.ControlStyles.AllPaintingInWmPaint
+					| swf.ControlStyles.DoubleBuffer
+					, BackColor.A != 255 );
+				base.OnParentBackColorChanged( e );
+			}
+		}
```
firda-cze: This eliminates flicker for content on drawable
...and generally transparent background

Source/Eto.WinForms/Forms/WindowsPanel.cs
```diff
@@ -113,7 +113,7 @@ public Control Content
 			}
 		}
 
-		protected virtual void SetContent(swf.Control contentControl)
+		protected virtual void SetContent(Control control, swf.Control contentControl)
```
firda-cze: Needed to detect unset background of content in `Drawable`

cwensley:
Hi @firda-cze,

Thanks for the updates, they look great!

As for the btw, I'm *not* suggesting for you to split up this PR, and I do realize that interdependent changes will need to be submitted at the same time.

My comments arose from the addition of the changes to add double buffering for the Grid/Tree, which seemed (to me) to be unrelated to the original intent of this PR which is to add transparency and proper layout handling to the Drawable.  This creates a moving target that makes it harder to review.  I will be more specific next time to hopefully clear up any misunderstanding.

Hopefully I did not offend you in any way, it was definitely not intended. Of course do what you feel is best for this or any PR.  All your work is very much appreciated!

Back to business, do you think this PR is ready to be merged?  I don't have anything specific on my end and it looks good.

I did notice that you're using C#6 syntax which is great!  I've been avoiding doing this up till now to support building the .NET 4.0 version with older versions of mono.  However, I would actually love to start using these features myself so we'll see if anyone runs into problems with them once it's merged (;

firda-cze:
It is ready. I may look into GTK2 drawables and transparency later, but cannot promise anything now.

The only thing I came across that does not currently work in mono (or those versions that my co-workes had on their Mac and Linux) was invariant interpolated string, So, I still rather avoid it (although it is very nice feature: `Invariant($"Format {me}")`).

BTW: I am currently very busy working on my **B#** (created repository for it here on github).

cwensley:
Just did some testing, and this looks awesome.  Thanks for the hard work getting this going!

I did notice that the transparent controls over drawable does indeed work on Mac, so I'll update the flags for that platform to reflect this.

Cheers!
Curtis.

firda-cze:
Good news! Happy to hear that :smile:

PR #473, opened 2016-03-16
Implement unhandled exceptions

ermshiperete:
This change allows to catch otherwise unhandled exceptions.
Currently this is implemented for GTK, WinForms and WPF.

---
This change is [Reviewable](https://reviewable.io/reviews/picoe/eto/473)

### DIFF
--- a/Source/Eto.Gtk/Forms/ApplicationHandler.cs
+++ b/Source/Eto.Gtk/Forms/ApplicationHandler.cs
@@ -25,6 +25,12 @@ namespace Eto.GtkSharp.Forms
 				SynchronizationContext.SetSynchronizationContext(new GtkSynchronizationContext());
 		}
 
+		void OnUnhandledException(GLib.UnhandledExceptionArgs e)
+		{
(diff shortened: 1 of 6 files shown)


cwensley:
Thanks for the PR!  Looks good, but a few points:

- Could you revert the changes to the .sln?  I see you've turned off compiling Direct2D and WinRT for debug builds, and possibly some other changes, but hard to tell.
- UnhandledExceptionEventArgs will need comments

cwensley:
Also, it should only subscribe the unhandled exception events in the AttachEvent() method in each of the handlers, so it is only wired up when it is used.

ermshiperete:
Done.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/picoe/eto/473#-:-KDPh7lLUIT7DlEEOrRN:-499146009)*

cwensley:
Looks excellent!! Thank you!

PR #479, opened 2016-03-31
Fix #478

### LINKED ISSUE #478: [GTK3] Resizable dialogs broken
From the example app:
untitled

cra0zy:


cwensley:
Thanks for the fix!  I can't remember what that code was supposed to do, but I'm guessing this may break something else such as minimum size, or some auto sizing behaviours.  I'll have to do some testing on that.

cra0zy:
I don't know why the code was there, but as for min/max size, the way you do it is using Gdk.Geometry: https://github.com/mono/MonoGame/blob/v3.5/Tools/Pipeline/Gtk/MainWindow.GUI.cs#L386

cwensley:
Sorry this took so long to merge.  Thanks for submitting the PR!

PR #482, opened 2016-04-04
[Gtk] Fix creating icon from fileName

cra0zy:
Notes
1) Faster and less code
2) The original method didn't work, kept throwing "unknown format"

### DIFF
--- a/Source/Eto.Gtk/Drawing/IconHandler.cs
+++ b/Source/Eto.Gtk/Drawing/IconHandler.cs
@@ -30,16 +30,8 @@ namespace Eto.GtkSharp.Drawing
 
 		public void Create(string fileName)
 		{
-			using (var fs = File.OpenRead(fileName))
-			using (var ms = new MemoryStream())
-			{
-				fs.CopyTo(ms);
-				ms.Position = 0;
-				Pixbuf = new Gdk.Pixbuf(ms);
-				Control = new Gtk.IconSet(Pixbuf);
-				ms.Position = 0;
-				CreateFrames(ms);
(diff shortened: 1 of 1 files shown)


cwensley:
Thanks for the PR!  Unfortunately, it appears as if it just reverts the update to support Icon.Frames.

What file are you loading?  is it an .ico or a different format of file?  Currently only .ico is supported across all platforms.

cra0zy:
.png

I looked at the examples and you seem to be loading .png images from resources in them, tho loading them from file crashes.

cra0zy:
@cwensley What would you suggest for implementing .png images? I meant where to implement it in Eto?

cwensley:
Right, that feature has only been implemented when loading from resources for now (where it loads @2x, etc files).  I would certainly like to see that supported from file, but hasn't been done yet.

The other platforms most likely will have the same problem as GTK at this point.

I've added issue #483 to track the feature.  If you'd like to try to tackle it let me know, otherwise I will try to get to it soon!

If you want to do this manually, you can compose an icon from many `Bitmap`s by passing an array of `IconFrame` objects to the `Icon` constructor.

Otherwise, use `new Bitmap(fileName)` if you just want to load a single .png directly.

Hope this helps!

cra0zy:
> Otherwise, use new Bitmap(fileName) if you just want to load a single .png directly.

Ok, I wouldn't have guessed that.

> Hope this helps!

As always you are very helpful, Thank You ^^

PS. MonoGame Pipeline Tool is being ported to Eto

PR #488, opened 2016-04-11
Fix Form position with Gtk3

cra0zy:
For `Control.SetPosition(Gtk.WindowPosition.Center);`:
 - on Gtk 2 it centers Window
 - on Gtk 3 it moves it to 0, 0, it doesn't even start it on the correct display, just moves it to 0, 0

This only disables the broken code on Gtk 3 so now the Window will at least start on the correct display.


cra0zy:
So, turns out this was a bug in Gtk 3 which got resolved few days ago... I upgraded the packages and centering works fine once again.

PR #495, opened 2016-04-17
Fix Context Menu Submenu and Clearing Items

cra0zy:
This fixes 2 things:
 1. Clearing items now actually removes submenu items instead of deleting the label of the main menuitem...
 2. Adds a ButtonRelease event to workaround Gtk ContextMenu bug that's been out for years. Basically what it does is makes you unable to select submenu items of context menu
   - https://bugzilla.gnome.org/show_bug.cgi?id=695488
   - http://stackoverflow.com/questions/5221326/submenu-item-does-not-call-function-with-working-solution


cwensley:
Excellent, thanks for the contribution!

PR #496, opened 2016-04-19
[GTK3] Implement New Color Chooser Dialog

cra0zy:
The old GtkColorSelection has been depricated in 3.4 and replaced by GtkColorChooser. This implements the new one and uses it in case of Gtk 3.4 or newer.

GtkColorSelection: https://developer.gnome.org/gtk3/stable/GtkColorSelectionDialog.html
GtkColorChooser: https://developer.gnome.org/gtk3/stable/GtkColorChooserDialog.html

### DIFF
--- a/Source/Eto.Gtk/Forms/ColorDialogHandler.cs
+++ b/Source/Eto.Gtk/Forms/ColorDialogHandler.cs
@@ -3,17 +3,38 @@ using Eto.Forms;
 
 namespace Eto.GtkSharp.Forms
 {
-	public class ColorDialogHandler : WidgetHandler<Gtk.ColorSelectionDialog, ColorDialog, ColorDialog.ICallback>, ColorDialog.IHandler
+    public class ColorDialogHandler : WidgetHandler<Gtk.Dialog, ColorDialog, ColorDialog.ICallback>, ColorDialog.IHandler
 	{
 		public ColorDialogHandler ()
 		{
-			Control = new Gtk.ColorSelectionDialog(string.Empty);
+#if GTK3
+            if (Gtk.Global.MinorVersion >= 4)
+                Control = new Gtk.Dialog(NativeMethods.gtk_color_chooser_dialog_new("Choose Color", IntPtr.Zero));
+            else
+#endif
+                Control = new Gtk.ColorSelectionDialog(string.Empty);
 			
 		}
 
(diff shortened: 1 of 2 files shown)


cwensley:
Great work, thank you for the PR!

Too bad we have to resort to using p/invoke ourselves here (again), but the new color picker is definitely nicer than the old one.

cra0zy:
> Too bad we have to resort to using p/invoke ourselves here (again), but the new color picker is definitely nicer than the old one.
 
There is also an option of using gtk-sharp 3.18 bindings available here: https://github.com/openmedicus/gtk-sharp

No Linux distribution provides them so they would need to be redistributed with Eto.Gtk3 nuget package (or use something like https://www.nuget.org/packages/gtk-sharp.Linux.3.14/ as a dependency).

cwensley:
Thanks @cra0zy for pointing that out, though I hesitate to add a dependency on a linux-specific nuget package since some people will want to develop or run it on other platforms..  

I also don't think I'll have the cycles to maintain gtk-sharp 3 for all distros (and all associated questions on why it doesn't run), I am really just biding time until someone steps up and publishes an 'official' gtk-sharp 3 for linux... I am really surprised it isn't more widely supported yet, but I guess there just isn't the need.

cra0zy:
You don't need to maintain it, just compile the repo and include .dll files from it in Eto.Gtk nuget package, let someone else maintain it.

This would also allow you to include a proper gtk3 sharp webview ( *caugh* https://github.com/openmedicus/webkitgtk3-sharp *caugh*), since no Linux distro includes it :(

> I am really surprised it isn't more widely supported yet, but I guess there just isn't the need.

The reason is that the monodevelop doesn't use GTK 3 :(

cra0zy:
Well I decided to maintain Gtk 3 sharp nuget packages:
 - https://www.nuget.org/packages/gtk-sharp3/
 - https://www.nuget.org/packages/webkitgtk3-sharp/

Sources for nuget packages are available:
 - https://github.com/cra0zy/gtk-sharp3-nuget
 - https://github.com/cra0zy/webkitgtk3-sharp-nuget

Just a side note I tried moving Eto to a newer Gtk 3 version, there was one problem:
 - https://github.com/openmedicus/gtk-sharp/issues/9

I'll look into fixing it over the weekend if no one else does.

PR #504, opened 2016-04-26
Set default button for file dialogs

cra0zy:
If this property is not set all the buttons will look the same. You can see the difference with Gtk buttons on the image bellow:

![]
(https://developer.gnome.org/gtk3/stable/colorchooser.png)


cwensley:
Excellent, thank you!

PR #506, opened 2016-05-01
Mac: Fix WebView redrawing and file upload issue

lubos:
Two commits:

1. Adding back constructor for `ApplicationHandler` on Mac. Without this constructor, WebView on Mac doesn't redraw itself properly when rendering HTML/CSS with fixed positioned elements.

2. It turns out when overriding `UIRunOpenPanelForFileButton` method in WebView on Xamarin.Mac, `resultListener` argument will never be `WebOpenPanelResultListener` type therefore file upload function in WebView will be broken. Xamarin.Mac has new helper method which should be used instead and it fixes the issue with file upload.

### DIFF
--- a/Source/Eto.Mac/Forms/ApplicationHandler.cs
+++ b/Source/Eto.Mac/Forms/ApplicationHandler.cs
@@ -48,6 +48,11 @@ namespace Eto.Mac.Forms
 
 		public bool AllowClosingMainForm { get; set; }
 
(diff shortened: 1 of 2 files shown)


cwensley:
Hi @lubos, thanks for submitting this issue.  It's very odd behaviour that redrawing has issues without keeping a reference to NSApplication, but at least that works!

PR #508, opened 2016-05-02
Fix ContextMenu

cra0zy:
It was possible for both events to occur, only one needs to occur.


cra0zy:
I think I can improve this, don't merge it yet.

cra0zy:
@cwensley Please go over this when you get the chance.

cwensley:
Thanks for the PR!  Looks good.

PR #511, opened 2016-05-12
[GTK3] Fix context menu... again...

cra0zy:
@cwensley  I am sorry for not noticing that I managed to brake something...

I added a deactivate event in Button release(because if I don't add that, something like recent menu won't be able to modify itself), but in doing so I managed to stop Activate event from ever occurring...


cwensley:
@cra0zy no worries, it happens (;  Thanks for submitting the update!

Issue #512, opened 2016-05-13
[Gtk] DropDown not reacting to mouse events


cra0zy:
Connecting to mousedown, mouseup, mousemove... events does nothing, I suspect that Eto may not be wrapping gtk combobox inside eventbox.

PR #513, opened 2016-05-14
Gtk Add Combobox and DropDown to Eventbox so mouse events occur

cra0zy:
Fixes: #512

For DropDown no mouse events worked.
For ComboBox only mouse move event worked.


cwensley:
To give some background, this gives a few benefits:

- Makes the handlers more straightforward, as Control is always the main content of the handler
- The ControlObject property of the Eto control will also return the Gtk.ComboBox, which is the most useful when trying to interact with the native control objects.

cra0zy:
Done.

cwensley:
AWESOME!!  Thanks, right under the line. (;

cra0zy:
Wow that was a fast response...

cwensley:
I'm on a rampage!  (;  That and I'm finishing up the 2.3 release.

cra0zy:
Ah, makes sense.

PR #517, opened 2016-05-23
GtkSharp 3.20 Update

cra0zy:
Native "glue" libs are available in x86 and x64 folders, these folders get copied to output directory by the nuget package.

I still need to test this out on Ubuntu and setup a way for an application to be able to select one of the folders above as the lib source.


cwensley:
Hi @cra0zy, looking like gtk#3 from nuget is getting closer to a reality.

Do you know if the 'glue' libs work for linux (.so), or are they windows only (.dll)?

cra0zy:
> Do you know if the 'glue' libs work for linux (.so), or are they windows only (.dll)?

Right now the nuget package only has Linux glue libs, I'll do an update soon adding Windows and Mac libs, and updating the config files so that they have proper libs from the start. I spent last 2 days setting up a proper Mac VM.

cra0zy:
Ok so...

For Windows MSYS was highly broken and I didn't get anywhere...

For Mac I need to wait for http://brewformulas.org/Gtkx3 to be 3.20, currently it's on 3.18.9 (every 0.2 is a stable release so it's literally the last stable version before 3.20).

cra0zy:
This just became way more confusing: https://blogs.gnome.org/desrt/2016/06/13/gtk-4-0-is-not-gtk-4/

cra0zy:
@cwensley Would it be Ok if I started submitting PRs to replace gtk-sharp for Gtk 3 with full native wrappers?

cwensley:
Hey @cra0zy, I would be okay with that, but we need to figure out a few things first.. Namely, what are the requirements for someone to run such an application?

Right now in ubuntu, one only has to install the "gtk-sharp3" package, and their app will run.  Would your changes mean that they would have to package the gtk-sharp3 assemblies with their app?  If so, it would be difficult to publish the app to the ubuntu as they will reject it immediately if it includes its own copy of platform-supplied assemblies.

If we can keep the ability to use the platform-supplied version(s) of gtk-sharp, then PR away (;

cwensley:
Hm, re-reading your comment..  do you mean that you'd like to get rid of the gtk-sharp & cairo-sharp dependencies altogether?  That would certainly be interesting. 

I'd welcome PR's to add that capability for sure, but I  wouldn't be able to help maintain it.  I would also suggest that it be a different platform implementation, keeping the existing gtk#2/3 platforms in place.

cra0zy:
Yes, the idea is to get rid of the dependency all together, I wouldn't mind maintaining it afterwards.

cwensley:
Excellent, that sounds good then!  I look forward to seeing what you come up with, and I'm sure others would be interested in this as well. (;

Issue #550, opened 2016-08-16
[WPF] Scrollable not sizing correctly (2.2 and 2.3 dev branch)


se5a:
This branch shows the issue well:
https://github.com/se5a/EtoTests/tree/ScrollableTestDevBranch/EtoApp1

To reproduce in the above:
run with WPF as startup project.
Click the Scrollable Test button
A new tab will be created with a scrollable and a number of items.
The scrollable is not sized correctly and the button is not visible at the bottom. 
doing the above with GTK2 as the startup will show what the scrollable should look like. 

Note: I *think* I recall testing this in an older version of the 2.3 dev branch some months ago and it appeared that the problem had been fixed, this may be a regression.

cwensley:
Now fixed!  Thanks for submitting the issue and the test app, it really helped ensure things were working right!

PR #551, opened 2016-08-17
Improved calculation of the PreferredSize for GroupBox in WPF.

yu-kopylov:
This commit should more or less fix the issue #550.

It is slightly inaccurate.
The calculated PreferredSize value for GroupBox is less than a pixel greater than it should be.
This inaccuracy is visible if there are a lot of GroupBoxes in a Scrollable.

The default GroupBox template can be extracted as described here: http://stackoverflow.com/questions/8825030/how-to-extract-default-control-template-in-visual-studio


cwensley:
Hey I've committed a fix for this in commit b30598642fab8623d24b825f46b8fb491aea1cb0, it was because Eto was measuring of the content of the GroupBox, not the entire GroupBox.  It should better handle any changes to the GroupBox style as well.

I'll close this one for now, but if it doesn't fix your issue please feel free to let me know and we'll re-visit.

PR #554, opened 2016-08-18
CefSharp control for Eto.Forms

cwensley:
Adds cefsharp-backed WebView support for WPF and WinForms platforms.

Also includes an experimental mechanism to more easily register additional platform handlers using assembly attributes, which is certainly not finalized.  The idea is that this should "automagically" register any assemblies that provide additional handlers just by referencing it.. though the problem so far is that the referenced assembly isn't actually referenced unless you use any of its code.

One idea might be to look at any assembly in the startup executable's folder, but that may cause other problems..


cwensley:
Should be moved to Eto.Toolkit: https://github.com/picoe/Eto.Toolkit/issues/8

Issue #462, opened 2016-02-10
TableLayout - how to add/remove rows at runtime


mennodeij:
As per title - I want to use TableLayout as a sort of list, to which rows can be added at runtime. 
The whole thing works (and I see that the #rows goes up), except that no changes are shown on screen.

cwensley:
Unfortunately at this time you can't dynamically add rows to the TableLayout.  I am however working on a major update to the layout infrastructure in Eto.Forms (see #248, #274, #276) that will allow for dynamic changes to the rows/columns of a table layout.

The only way for you to do this at this moment is to recreate the TableLayout entirely when you want to change its rows/cells.

Alternatively, the StackLayout does support adding/removing items dynamically.  However, it also uses the TableLayout so it is the same as doing the above.

mennodeij:
Ah, yes. Re-creating the table-layout works fine, thanks for the suggestion.

PR #556, opened 2016-08-18
Layout Rewrite

cwensley:
This PR is the initial layout re-write to move layout logic into Eto.Forms instead of relying on each platform control to provide this functionality (which each have hacks to make them behave similarly).

Ideally we can get rid of the hacks, and just make sure each control supports a Measure() and Layout() pass, where the Measure() would determine its preferred size based on the container, and Layout() would position its child controls.  This would greatly simplify the layout logic, make it easier to add other platforms, and also make it much easier to add additional layout options, such as a FlowLayout (which is included in this PR), etc.

### DIFF
--- a/Source/Eto.Gtk/Eto.Gtk2 - net45.csproj
+++ b/Source/Eto.Gtk/Eto.Gtk2 - net45.csproj
@@ -260,6 +260,9 @@
     <Compile Include="..\Shared\BaseBitmapData.cs">
       <Link>Drawing\BaseBitmapData.cs</Link>
     </Compile>
+    <Compile Include="Forms\CustomLayoutHandler.cs">
+      <SubType>Code</SubType>
+    </Compile>
   </ItemGroup>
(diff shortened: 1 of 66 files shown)


cra0zy:
I know Gtk doesn't like to be told specific sizes and also trying so set specific pixel sizes will probably break with Gtk when the user switches theme, since each element has it's size info stored with the theme. You might want to consider using Gtk layout way as a base, it might be a lot easier.

cwensley:
@cra0zy, the idea is to ask the platform which size each control should be, so it should still adjust based on the theme.  Each control should be able to return its 'desired size', which will be platform specific.  I'll look at the Gtk layout as a base, which could probably hook into this mechanism.

cra0zy:
> Each control should be able to return its 'desired size', which will be platform specific.

If the user changes theme while the app is running, the starting 'desired size' size will be wrong. Not sure if gtk has a theme changed event, but if it does, you will need to reset sizes when it occurs.

cwensley:
This should fix #462 when implemented

PR #561, opened 2016-09-05
Fix RadioCommand NullException when using with Menu

cra0zy:
The error was happening because you were trying to connect events before creating the main object.


cwensley:
Nice, thanks for the fix!

PR #564, opened 2016-09-08
Add secure passwordbox

mennodeij:
This PR adds a more secure implementation of the PasswordBox using a SecureString as backing store. 
I have tested this with WPF and Gtk3 - I don't have access to other platforms.

### DIFF
--- a/Source/Eto/Eto.csproj
+++ b/Source/Eto/Eto.csproj
@@ -68,6 +68,7 @@
     <Compile Include="Forms\Controls\SearchBox.cs">
       <SubType>Eto.Forms</SubType>
     </Compile>
+    <Compile Include="Forms\Controls\SecurePasswordBox.cs" />
     <Compile Include="Forms\Controls\TextInputEventArgs.cs" />
     <Compile Include="Forms\Controls\ThemedContainerHandler.cs" />
     <Compile Include="Forms\Controls\ThemedControlHandler.cs" />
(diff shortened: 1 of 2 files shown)


mennodeij:
Hmm... Maybe I'm missing something here, but the AppVeyor build did _not_ build my additions. It checked out https://github.com/picoe/Eto.git and built that.

cwensley:
Hey @mennodeij, thanks for the PR!

Unfortunately, it's usefulness on mono is fairly limited, as [SecureString isn't actually secure](https://github.com/mono/mono/blob/master/mcs/class/corlib/System.Security/SecureString.cs#L253).  

That being said, it might be worth investigating whether the existing `PasswordBox` can be updated to add a `SecureString` property instead - on WPF at least it has a `SecurePassword` property so this can be directly accessed and will indeed be fully secure. 

What do you think?

mennodeij:
Would that mean that on WPF the password is only stored securely?

cwensley:
@mennodeij right, it wouldn't be secure on any platform other than WPF or WinForms.  

Even then, the security of it all is suspect as eventually you translate it to a normal string to actually use it.  Also, on [msdn](https://msdn.microsoft.com/en-us/library/system.security.securestring(v=vs.110).aspx) it states: 
> On the Windows operating system, the contents of a SecureString instance's internal character array is encrypted. However, whether because of missing APIs or key management issues, encryption is not available on all platforms. Because of this, SecureString is available on Desktop (Windows only), and not on .NET Core.

So, it's not really useful as it isn't cross platform at all.

I'll close this now but if you feel like it should still be in Eto, let me know!

PR #557, opened 2016-08-24
Fix Gtk2/Gtk3 button menu event firing for Ubuntu

jaebird:
Ubuntu application menus act slightly differently than standard Gtk menus. This one line change fixes menu items not firing to the event handlers in client code.

PR #569, opened 2016-09-29
gtk3 text view uses the wrong default text color, gtk entry is correct.

cra0zy:
Extracted from: #557


cwensley:
Thanks for separating that out, great work!

PR #570, opened 2016-10-07
Fix removing controls from a TableLayout

cwensley:
- Also detach a control when it is removed from the logical parent, so that when you remove items from StackLayout/etc it will remove it from the visual tree immediately.


cwensley:
Apparently, due to naming conflict with method in the base class.  **Note to self:** always deal with warnings.

Issue #566, opened 2016-09-26
Suggestion: A few common used controls


claunia:
May I suggest, a few controls/widgets that I've seen used a in a lot of places, yet it seems everyone has to custom code them.

A pixel map (not an image) that draws a Color[] as a series of squares and can optionally put a border separating them.
Example:
![](http://imag.malavida.com/mvimgbig/download-fs/nero-discspeed-13989-3.jpg)

Pie charts, filled to the center
![](https://camo.githubusercontent.com/9311e90656dde15b32058a6419e9e8cdfc64dea7/687474703a2f2f692e696d6775722e636f6d2f4e4e6b43574e7a2e6a7067)

and not
![](https://camo.githubusercontent.com/64691a56adb5a2536f9ece2d174f645dfb3f8a7f/687474703a2f2f692e696d6775722e636f6d2f4555664f3870762e706e67)

Circular progress "bars"
![](http://demo.qodeinteractive.com/bridge-help/assets/images/image_74.jpg)

And why not other kind of plottings
![](http://sibeeshpassion.com/wp-content/uploads/2016/03/Chart-Widgets-With-Server-Side-Data-In-MVC-Using-Angular-JS-And-Web-API-Output-1024x875.png)

Gauges
![](http://2.bp.blogspot.com/-13rb4kB7UAc/UDzUi5nsAZI/AAAAAAAAASE/NTviTnXLMCY/s1600/AGaugeApp.jpg)

System tray / OSX Menubar icon.

cwensley:
Thanks for the suggestion(s), though it's quite hard to track an issue with so many requests together.  It is typically better to split them up so they can get different priority, etc.  Anyway, here's some feedback that might help others, if they decide to look at doing any of this:

For charts/graphs, I suggest looking at OxyPlot which @LorenVS made Mac/WPF wrappers for [here](https://github.com/LorenVS/Eto.OxyPlot), or you can fix up the OxyPlot.Eto port [here](https://github.com/rconde01/oxyplot/tree/develop/Source/OxyPlot.Eto.Forms_NET40) to support all of the platforms Eto supports.

As for the the pixel map, it'd be pretty dead simple to make your own based on the Drawable control.

I'm not sure what's available for Gauges, but it could be a lot of work if starting from scratch.

Cheers!
Curtis.

claunia:
@cwensley yeah I know but it's more like a list of suggestions that can then be split in separate issues for implementation. I though it was a good idea.

[...]

PR #586, opened 2016-10-26
Add helpers to create a Bitmap from an enumeration of Color or int32 …

cwensley:
…values

Add helpers to BitmapData to set/get pixel data as Color values


claunia:
A step forward to one of my suggestions on #566.

Cheers!

cwensley:
Yep, that did influence adding this feature, so thanks again for the input!

claunia:
@cwensley only missing the borders, but a lot easier now than before! I still find stunning how many software uses that yet they need to custom implement it.

PR #595, opened 2016-11-07
[Gtk] Fix button order for YesNoCancel dialog

cra0zy:
The standard order for Gtk YesNoCancel dialog should be "No|Cancel|Yes", the current one was "No|Yes|Cancel". This type of dialog is easily seen in Gedit.


cwensley:
Looks great!

PR #596, opened 2016-11-07
[Gtk] Allow combobox to be lower size than it's longest item

cra0zy:
This PR allows for the combobox to be lower width than it's longest item. In case the item cannot fit in the combobox it will have "..." at the end.

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/DropDownHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/DropDownHandler.cs
@@ -64,6 +64,20 @@ namespace Eto.GtkSharp.Forms.Controls
 			}
 		}
 
+		public override Size Size
+		{
+			get { return base.Size; }
+			set
+			{
+				if (value.Width == -1)
+					text.Ellipsize = Pango.EllipsizeMode.None;
+				else
+					text.Ellipsize = Pango.EllipsizeMode.End;
+
+				base.Size = value;
+			}
+		}
+
 		public virtual int SelectedIndex
 		{
 			get { return Control.Active; }


cra0zy:
I thought of a better way to do this.. `text.Ellipsize = Pango.EllipsizeMode.End;` makes it so if the combobox doesn't have enough space it adds `...` to the end, but the negative side is that combobox no longer has any width request so it becomes very small, so the solution is to set some default width and allow the end user to override it with `Width` property of the control.

Source/Eto.Gtk/Forms/Controls/DropDownHandler.cs
```diff
@@ -35,6 +36,7 @@ protected override void Initialize()
 		{
 			Create();
 			container = new Gtk.EventBox();
+			container.WidthRequest = 150;
```
cra0zy: Kind of, if it's placed inside a layout it will expand if the layout requests it to expand, but it will not try to set it's own minimum size to the longest item size.

cra0zy:
Sometimes I feel dumb, and this is one of those cases. Anyway I know how to solve this without the Width problem, I'll submit a PR tomorrow morning (I actually went to bed/try to sleep 30 minutes ago...)

cra0zy:
@cwensley Updated :)

Source/Eto.Gtk/Forms/Controls/DropDownHandler.cs
```diff
@@ -64,6 +64,18 @@ public virtual void HandleChanged(object sender, EventArgs e)
 			}
 		}
 
+		public override Size Size
+		{
+			get { return base.Size; }
+			set
+			{
+				if (value.Width != Size.Width)
+					text.Ellipsize = Pango.EllipsizeMode.End;
```
cra0zy: I did not know that you used `-1` for setting auto size... I'll change the code as soon as I can.

cwensley:
Beauty, thanks!

PR #606, opened 2016-11-22
AboutDialog Base

cra0zy:
Stuff done:
 - Adds `AboutDialog` base class
 - Adds `AboutDialogHandler` for Gtk 2 and 3 backends (https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html)
 - Adds `ThemedAboutDialogHandler` for the other backends
 - Adds a section for `AboutDialog` in the Test App


cra0zy:
@cwensley This is ready for merging.

cwensley:
Looks really good, thanks again!  A few minor thoughts:

- It might be a good idea to add an (optional) `Assembly` parameter to the constructor so you can supply which assembly to get the default information from instead of always the caller's assembly
- I'm not sure if this matters, but should there be a way to construct the dialog _without_ supplying information from an assembly?
- I tend not to use the `private` keyword as it is the default

cra0zy:
I've updated the PR with requested changes.

> I'm not sure if this matters, but should there be a way to construct the dialog without supplying information from an assembly?

I doubt there is a reason to make one.

cwensley:
Awesome, thanks for the quick turnaround!

cwensley:
> tho I tend to use MS rules so I do forget some stuff from time to time.

That is fair enough, I am warming to the idea to go along with MS rules, it's just hard to re-train yourself to do things differently.  Using their roslyn tool to verify code formatting during CI builds might be interesting

PR #614, opened 2016-11-28
[Gtk] Fix SpinButton not having enough space to be allocated properly

cra0zy:


cwensley:
From what I recall, without this, and if your NumericUpDown is auto sized, its width would be huge.. is this not the case anymore?

cra0zy:
It isn't, and with the code that you've been using you have been cutting it half way and not shrinking it.

PR #636, opened 2016-12-18
Wpf: Use native Scrollable content expansion

TripleWhy:
This prevents an issue demonstrated here: https://github.com/TripleWhy/EtoErrorDemo/tree/master.
When executed on Windows 10 with display scaling enabled, eto.wpf runs into an infinite loop that alternates the visibility of the scroll bars. This PR fixes that.


TripleWhy:
Well, apparently this breaks Eto.Test.Wpf > Controls >TextArea and causes it to expand its width indefinitely instead of its height.

Issue #644, opened 2016-12-30
Splitter.Panel.Visible not working on WPF


TomQv:
When setting a Splitter.Panel.Visible=false, the other panel should resize to fill the whole space of the splitter. This works with WinForms and OSX. With WPF the panel just hides and the other one is not resizing.

cwensley:
Hm, I thought I got this working.. have you tried the develop branch?  I added a test specifically for this case so it should be working.

TomQv:
ye, develop branch. It was fixed for Mac, but WPF I never used until now-

cwensley:
Could you make a little test case for your issue?  The test [here](https://github.com/picoe/Eto/blob/develop/Source/Eto.Test/Eto.Test/Sections/Controls/SplitterSection.cs#L452), which sets Splitter.Panel1/2.Visible works for me.

TomQv:
your sample is working, but my usecase is more complex. In our app we need a UI similar to the IDE of Visual Studio with hidable toolwindows at the side and multiple documents in the middle. I'm still not satisfied with what I've reached so far, but I also don't have time for that currently. 
btw. how about multiple toolbars?

cwensley:
Hm, no problem.  I won't be able to fix this without any further detail on why it doesn't work for you, perhaps some hints on what kind of things you do might help.

Multiple toolbars is tracked with issue #339.

TomQv:
The [old sample](https://github.com/picoe/Eto/issues/500#issuecomment-214540995) I did for the same problem with Mac is showing the issue on Wpf.
Hope this helps.

cwensley:
@TomQv thanks for pointing to the sample, but it has too many options for me to know what it is doing.  

Could you tell me what exactly I should be clicking on in the sample and what you expect that is not happening?

TomQv:
sorry ;-)
there is two splitterpanels in the middle. The left one is titled "_xp2" the right one is "_mapWindow1"
With the View-Menu you can change visibility:
- Map:       "_xp2" hidden, "_mapWindow1" visible
- Xplorer:   "_xp2" visible, "_mapWindow1" hidden
- Xplore-Map:   "_xp2" visible, "_mapWindow1" visible
When one gets hidden, the other one must resize to fill the whole space, which is not happening on Wpf, but on WinForms and Osx.
There is also three more splitterpanels which you can hide/view with the View-Menu - Tools left,right,bottom. Same issue.

Tom

cwensley:
[...]

PR #648, opened 2017-01-03
Add additional tests for Splitter hiding

cwensley:
- Test when replacing Panel1 or Panel2 with a new control with different visibility, or null.
- WinForms: Fix setting Panel1/Panel2 to new control when its initial Visible state is set.
- Fix Unbind() by removing the item from Properties, instead of setting the value to null.


cwensley:
Additional tests for issue #644

PR #653, opened 2017-01-05
Implement FontPicker

cra0zy:


Source/Eto.Gtk/Forms/Controls/FontPickerHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+		{
+			Control = new Gtk.FontButton();
+		}
+
+		public override void AttachEvent(string id)
+		{
+			switch (id)
+			{
+				case FontPicker.ValueChangedEvent:
+					Control.FontSet += (sender, e) => Callback.OnValueChanged(Widget, EventArgs.Empty);
```
cwensley: This will cause a leak unless it uses the Connector pattern.

cwensley:
This is awesome!  Great work on these.

Source/Eto.Gtk/Forms/Controls/FontPickerHandler.cs
cra0zy: You might need to fix up some stuff: https://github.com/picoe/Eto/blob/develop/Source/Eto/Forms/ThemedControls/ThemedTextStepperHandler.cs#L228

cra0zy: Where does the leak occur and why?

cwensley: That is different, as it's a Themed handler so it uses Eto controls.  

When subscribing to Gtk control events, this causes the Gtk control itself to never be GC'd.  If you add a finalizer to FontPicker it'll never get called even after many GC's.

This might be fixed in Gtk3, but at least this is how it worked with Gtk2 controls last I checked.

cwensley:
Excellent!  Thanks yet again. (;

PR #661, opened 2017-01-11
[Gtk] Fix scrollable scroll event

cra0zy:
ScrollEvent is not actually a scroll event for ScrolledWindow in Gtk :) It's a general purpose check if mouse wheel moved.

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/ScrollableHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/ScrollableHandler.cs
@@ -108,10 +108,9 @@ namespace Eto.GtkSharp.Forms.Controls
 				Handler.Callback.OnSizeChanged(Handler.Widget, EventArgs.Empty);
 			}
 
-			public void HandleScrollableScrollEvent(object o, Gtk.ScrollEventArgs args)
+			public void HandleScrollableScrollEvent(object sender, EventArgs e)
 			{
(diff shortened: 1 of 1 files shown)


Source/Eto.Gtk/Forms/Controls/ScrollableHandler.cs
```diff
@@ -121,7 +120,8 @@ public override void AttachEvent(string id)
 			{
 				case Scrollable.ScrollEvent:
 					Control.Events |= Gdk.EventMask.ScrollMask;
-					Control.ScrollEvent += Connector.HandleScrollableScrollEvent;
+					Control.Vadjustment.ValueChanged += (sender, e) => Connector.HandleScrollableScrollEvent(ScrollPosition);
```
cwensley: I'm pretty sure this will still cause a memory leak as it creates a hard reference to the handler (similar to calling the Callback directly).  It should be something like this, changing the signature of `HandleScrollableScrollEvent` to match:
`Control.Vadjustment.ValueChanged += Connector.HandleScrollableScrollEvent;`

cra0zy: Ups, forgot about the memory leak issue, sorry.

cwensley:
Wow quick response.  Thanks for the fix!!

PR #664, opened 2017-01-12
[Gtk] Allow textbox to be smaller size

cra0zy:


cwensley:
Thanks!

cra0zy:
@cwensley Quick question, I'm planning on doing a bit of rework on the dialog class, that is removing the pointless `DefaultButton` and `AbortButton` properties and instead introducing a new method `void AddDialogButton(Button button, bool positive, bool defaultbutton)` that would also add button to the correct place in the dialog. What do you think on such a change?

cra0zy:
I know what the properties actually do, I just call them pointless because they do next to nothing.

cwensley:
@cra0zy they don't do nothing; the AbortButton is called on a dialog when you press the escape key, and the DefaultButton is called when you press enter.  On some platforms (e.g. macOS), it also gives them a different appearance (the DefaultButton is blue).

If you want a dialog with buttons in the standard/correct place (ala GTK since no other platform has this), then a Dialog subclass could be added that gives this functionality.  It would be ideal if instead of methods like `AddDialogButton()`, which are not very xaml or initializer pattern friendly, to have a collection of buttons that you can add to.  

Also, gitter is a great place for questions (;

cra0zy:
GitHub was already opened :P

cra0zy:
> If you want a dialog with buttons in the standard/correct place (ala GTK since no other platform has this), then a Dialog subclass

On Linux the positive response should be on the right, while on Windows it should be on the Left, what I want to add is a way for the buttons to automatically get correct positions.

This would also allow us to implement HeaderBar on Gtk Dialogs.

PR #666, opened 2017-01-12
Added PositiveButtons and NegativeButtons to Dialog

cra0zy:
This is only implemented for Gtk 2 and 3 right now.


cwensley:
Just so you know, I haven't forgotten about this.  I ran into a problem using Eto controls inside the handler, as currently how containers work is that the Eto implementation determines whether the control is logical or native.

This means that we can only have all platforms use native children, or all use non-native (Eto) children.  This is not ideal, and the handler itself should decide (or at least have control over) how to hook up the children. I'm still contemplating different ways we can accomplish this, which when done will make this easy to finish up.

Alternatively, we can just implement all handlers using native children.  This is not the end of the world as it isn't very complex logic to position the buttons on each platform. 

I'm thinking of taking the second path for now, and possibly address this with the 3.0 release as the handler implementations might all change anyway.

cra0zy:
> I'm thinking of taking the second path for now, and possibly address this with the 3.0 release as the handler implementations might all change anyway.

Sounds good to me.

cra0zy:
@cwensley WPF and WinForms implemented, all that is left is your Mac side :)

cra0zy:
Well... I need this today... so I guess I'll do the Mac backend as well...

cra0zy:
Actually scratch that, I'll just remove this feature from the secondary PR, I have no idea how Xamarin.Mac layouts work...

PR #671, opened 2017-01-20
Alternate method to hook up properties & events

cwensley:
This is an experimental feature, trying to simplify the creation of new controls.  In most cases this may even eliminate the need to declare an IHandler interface for the control.

This will probably be done for the 3.x series (which is a ways off)

### Benefits:
1. If done properly, we can make it so only a single handler instance is created for ALL child objects instead of one per object reducing the number of instantiated objects substantially
2. Simplifies not only the declaration of the control but could also simplify platform implementations by eliminating the need for the `Connector` in GTK (due to using static methods to wire up event handlers)
3. Removes the need for the (excessive) amount of code required for `ICallback`/`Callback` implementations
4. Removes the need for `EventManager.Register()` in the static constructors
5. Removes the need for `IHandler` in cases where there are no methods and only properties and/or events.
6. Change is good

### Drawbacks:
1. Completely breaks any custom handlers, custom controls, or styles for native controls done through the handler
2. If more properties are stored in the `Widget.Properties`, accessing them may be slower (though this can be mitigated by pushing more functionality into subclasses of the native controls)
2. Change is bad

### Incomplete:
- [ ] Need to figure out how to style native controls, e.g. `Style.Add<SomeHandler>(h => h.Control.NativeProperty = blah);` won't work
- [ ] Other stuff I haven't thought of yet

### DIFF
--- a/Source/Eto.Mac/Forms/Controls/StepperHandler.cs
+++ b/Source/Eto.Mac/Forms/Controls/StepperHandler.cs
@@ -32,6 +32,21 @@ using CGPoint = System.Drawing.PointF;
 
 namespace Eto.Mac.Forms.Controls
 {
+	public class MacViewHandler2<TWidget, TControl> : WidgetHandler2<TWidget, TControl>, IMacControlHandler2
+		where TWidget : Control
+		where TControl : NSView, new()
+	{
+		public virtual NSView GetContainerControl(Widget widget)
+		{
+			return GetControl((TWidget)widget);
+		}
+
+		public SizeF GetPreferredSize(Widget widget, SizeF availableSize)
+		{
+			return SizeF.Empty;
+		}
+	}
+
 	public class StepperHandler : MacControl<NSStepper, Stepper, Stepper.ICallback>, Stepper.IHandler
 	{
 		public class EtoStepper : NSStepper, IMacControl
@@ -49,17 +64,9 @@ namespace Eto.Mac.Forms.Controls
 			};
 		}
 
-		static object ValidDirection_Key = new object();
-
-		public StepperValidDirections ValidDirection
-		{
-			get { return Widget.Properties.Get(ValidDirection_Key, StepperValidDirections.Both); }
-			set { Widget.Properties.Set(ValidDirection_Key, value, UpdateState, StepperValidDirections.Both); }
-		}
-
 		void UpdateState()
 		{
-			switch (ValidDirection)
+			switch (Widget.ValidDirection)
 			{
 				case StepperValidDirections.Both:
 					Control.ValueWraps = true;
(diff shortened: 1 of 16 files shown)


cwensley:
This allows us to change something like this (e.g. the Stepper control):

```c#
[Handler(typeof(IHandler))]
public class Stepper : Control
{
	new IHandler Handler { get { return (IHandler)bqase.Handler; } }
	
	static Stepper()
	{
		RegisterEvent<Stepper>(c => c.OnStep(null));
	}

	public const string StepEvent = "Stepper.Step";

	public event EventHandler<StepperEventArgs> Step
	{
		add { Properties.AddHandlerEvent(StepEvent, value); }
		remove { Properties.RemoveEvent(StepEvent, value); }
	}

	protected void OnStep(StepperEventArgs e)
	{
		Properties.TriggerEvent(StepEvent, this, e);
	}

	[DefaultValue(StepperValidDirections.Both)] 
	public StepperValidDirections ValidDirection
	{
		get { return Handler.ValidDirection; }
		set { Handler.ValidDirection = value; }
	}

	ICallback callback = new Callback();

	protected override object GetCallback() => callback;

	public new interface ICallback : Control.ICallback
	{
		void OnStep(Stepper widget, StepperEventArgs e);
	}

	protected new class Callback : Control.Callback, ICallback
	{
		public void OnStep(Stepper widget, StepperEventArgs e)
		{
			widget.Platform.Invoke(() => widget.OnStep(e));
		}
	}

	public new interface IHandler : Control.IHandler
	{
		StepperValidDirections ValidDirection { get; set; }
	}
}
```

To this:

```c#
public class Stepper : Control
{
	public static DependencyEvent<Stepper, StepperEventArgs> StepEvent = new DependencyEvent<Stepper, StepperEventArgs>((c, e) => c.OnStep(e));

	public event EventHandler<StepperEventArgs> Step
	{
		add { Properties.AddEvent(StepEvent, value); }
		remove { Properties.RemoveEvent(StepEvent, value); }
	}

	protected void OnStep(StepperEventArgs e) => Properties.TriggerEvent(StepEvent, this, e);


	public static DependencyProperty<Stepper, StepperValidDirections> ValidDirectionProperty = new DependencyProperty<Stepper, StepperValidDirections>(StepperValidDirections.Both);

	[DefaultValue(StepperValidDirections.Both)] 
	public StepperValidDirections ValidDirection
	{
		get { return Properties.Get(ValidDirectionProperty); }
		set { Properties.Set(ValidDirectionProperty, value); }
	}
}
```

cwensley:
### Benefits:
1. If done properly, we can make it so only a single handler instance is created for ALL child objects instead of one per object reducing the number of instantiated objects substantially
2. Simplifies not only the declaration of the control but could also simplify platform implementations by eliminating the need for the `Connector` in GTK (due to using static methods to wire up event handlers)
3. Removes the need for the (excessive) amount of code required for `ICallback`/`Callback` implementations
4. Removes the need for `EventManager.Register()` in the static constructors
5. Removes the need for `IHandler` in cases where there are no methods and only properties and/or events.
6. Change is good

### Drawbacks:
1. Completely breaks any custom handlers, custom controls, or styles for native controls done through the handler
2. If more properties are stored in the `Widget.Properties`, accessing them may be slower (though this can be mitigated by pushing more functionality into subclasses of the native controls)
2. Change is bad

### Incomplete:
- [ ] Need to figure out how to style native controls, e.g. `Style.Add<SomeHandler>(h => h.Control.NativeProperty = blah);` won't work
- [ ] Other stuff I haven't thought of yet

Issue #493, opened 2016-04-15
TreeGridView no way to add just one item


cra0zy:
Currently from what I've seen, the only way of adding 1 value to treeGridView is by refreshinh the entire DataStore... which when you have a lot of values isn't that lightweight.

cwensley:
Thanks for reporting!  with the `TreeView`, you can use the `RefreshItem` on the parent node to invalidate its children collection.  However you are correct in that the `TreeGridView` does not support this very well.

This is related to #240 where it should support hooking into MVVM and collection changes to automatically update the UI when the model changes.

cra0zy:
Fixed in #672

Issue #451, opened 2016-01-30
Gtk TreeGridView SelectedItems only registers first item


cra0zy:
With Gtk backed, and TreeGridView with multiselect enabled, I am unable to get selected items. The SelectedItems property is in most cases empty, and the only moment that it's not empty is if the tree root node is selected.

cwensley:
Thanks for reporting the issue!  I'll take a look.

cra0zy:
The error is somewhere in either GetRowIndexOfPath or DataStore.

cra0zy:
@cwensley Why didn't you make SelectedItems do the same thing as SelectedItem?

cwensley:
@cra0zy, it looks like `TreeGridView.SelectedItems` was never implemented properly and was just copied from `GridView` four years ago.  It will need a proper implementation, and to be efficient the implementation should probably be moved to the handler.

cra0zy:
> @cra0zy, it looks like TreeGridView.SelectedItems was never implemented properly and was just copied from GridView four years ago. It will need a proper implementation, and to be efficient the implementation should probably be moved to the handler.

That is something that's much easier to do. I can do it for few platforms, what should I do about the rest? Add a NotImplementedException?

PR #672, opened 2017-01-21
TreeGridView updates

cwensley:
This is a work in progress, and will hopefully fix a ton of issues with the TreeGridView.  Not full MVVM yet, but one step closer to getting that implemented.

It adds some apis:
- `TreeGridView.GetCellAt(PointF location)` - get the column & item at a specified point
- `TreeGridView.ReloadData()` - reload all data, keeping selection
- `TreeGridView.ReloadItem(ITreeGridItem item)` - reload the specified item and its children (helps with #493, but still not a full fix yet)
- `Grid.Border` - specify the type of border, or none (similar to Scrollable)

Fixes some things:
- Make mouse up/down events consistent and usable, combined with `GetCellAt()` should be very useful
- Fix `TreeGridView.SelectedItems` when allowing multiple selection (#451)
- Performance issues on some platforms
- Expanding children (if Expanded = true) when the parent node is expanded

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/GridHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/GridHandler.cs
@@ -21,6 +21,8 @@ namespace Eto.GtkSharp.Forms.Controls
 
 		protected Dictionary<int, int> ColumnMap { get { return columnMap; } }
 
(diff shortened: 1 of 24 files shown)


cra0zy:
So exciting :D

PR #676, opened 2017-01-26
Gtk2: fix gtk splitter to work the old way as it’s causing too many crashes with Eto.Test.

cwensley:
- Add SplitterHandler.ShrinkContentsToFit property so you can get the new functionality with a style.
- Gtk3 will retain the new functionality as it doesn’t crash.
- Fix leak when recreating the paned control by unhooking the Realized event

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/SplitterHandler.cs
@@ -13,6 +13,32 @@ namespace Eto.GtkSharp.Forms.Controls
 		int? position;
 		double relative = double.NaN;
 		int suppressSplitterMoved;
+		bool shrinkContentsToFit = true;
+
+		/// <summary>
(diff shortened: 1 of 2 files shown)


cwensley:
@cra0zy just a heads up.  I'm still encountering numerous crashing issues with Gtk2 in the Eto.Test.  A simple repro is going to **Behaviours > Badge Label** then resizing the form smaller than the content of the right split pane.

I've only set it to use the "old" way of sizing for Gtk2, but you can control it using the `SplitterHandler. ShrinkContentsToFit` property in a style either way.

cra0zy:
Is there a way to set splitter min and max position?

cwensley:
@cra0zy no, but sounds like a good idea for an enhancement.  Might be doable just by trapping the `PositionChanged` event and then ensuring the position is in the desired range.

cra0zy:
Then change the bool back to true, I'll get the min / max position implemented next week (I have some exams these next few days so no time :/)

PR #683, opened 2017-02-01
Add DropDown.ItemImageBinding

### LINKED ISSUE #61: Combobox doesn't show images when ImageListItem is used
e.g.
fcombobox.Items.Add (new ImageListItem { Text = "Item " + list.Items.Count, Image = myImage});

cwensley:
- Allows adding images to a DropDown using ImageListItem or your own custom POCO and binding.  Fixes #61
- ListBox now allows for differently sized images (not just 16 logical pixels high)
- Supports differently sized images (except WinForms), and macOS does not grow the control height so it overflows out of the control when selected.
- Wpf: streamline conversion of image to wpf
- Added Icon.WithSize() to return a copy of the Icon that will render at the specified logical size


cwensley:
.. also fixes regression in DropDown sizing in WPF when you explicitly set a size.

PR #693, opened 2017-02-02
[Gtk] Fix TreeGridView Multiselect

cra0zy:
In short, Gtk default treeview multiselect is extremely broken since right click does the same things as left click so you actually need a bit of code to make it act like the VisualStudio/XamarinStudio/MonoDevelop solution explorer treeview acts.

### DIFF
--- a/Source/Eto.Gtk/Forms/GtkControl.cs
+++ b/Source/Eto.Gtk/Forms/GtkControl.cs
@@ -451,7 +451,7 @@ namespace Eto.GtkSharp.Forms
 			}
 
 			[GLib.ConnectBefore]
-			public void HandleButtonPressEvent(object sender, Gtk.ButtonPressEventArgs args)
+			public virtual void HandleButtonPressEvent(object sender, Gtk.ButtonPressEventArgs args)
 			{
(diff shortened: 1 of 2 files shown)


Source/Eto.Gtk/Forms/GtkControl.cs
```diff
@@ -378,6 +378,8 @@ protected override WeakConnector CreateConnector()
 		/// </summary>
 		protected class GtkControlConnector : WeakConnector
 		{
+			protected bool buttonPressRetVal = true;
+
 			new GtkControl<TControl, TWidget, TCallback> Handler { get { return (GtkControl<TControl, TWidget, TCallback>)base.Handler; } }
 
 			public void HandleScrollEvent(object o, Gtk.ScrollEventArgs args)
```
cwensley: It might make sense to set this as virtual and override it from `GridConnector`, that way we don't need the extra protected bool for all controls.

cra0zy: Done.

cwensley:
Awesome!  Thanks for the fix.

PR #694, opened 2017-02-02
Splitter Min Width API

cra0zy:


cwensley:
Awesome!  Looks great.

A few minor things (I hope you don't mind my feedback!).  I'll comment inline.

Source/Eto/Forms/Controls/Splitter.cs
```diff
@@ -34,6 +34,9 @@ public enum SplitterFixedPanel
 	[Handler(typeof(Splitter.IHandler))]
 	public class Splitter : Container
 	{
+		bool shown;
+		int panel1MinWidth, panel2MinWidth = -1;
```
cwensley: The default values for the minimum width should probably be 0 instead of -1, as -1 doesn't have any meaning for a minimum size.  Also, these may be better named as `Panel1MinimumSize` and `Panel2MinimumSize`, as I'd like to refrain from using abbreviated names and also 'Width' does not apply if it is vertical.

Source/Eto/Forms/Controls/Splitter.cs
```diff
@@ -87,11 +90,43 @@ protected virtual void OnPositionChanged(EventArgs e)
...
+		}
+
 		static Splitter()
 		{
 			EventLookup.Register<Splitter>(c => c.OnPositionChanged(null), Splitter.PositionChangedEvent);
 		}
 
+		private void EnsurePosition()
+		{
+			if (Width <= 0)
```
cwensley: This method will have to be updated to deal with both horizontal and vertical orientations.

Source/Eto/Forms/Controls/Splitter.cs
```diff
@@ -87,11 +90,43 @@ protected virtual void OnPositionChanged(EventArgs e)
...
+			panel2MinWidth = -1;
+
+			Shown += Splitter_Shown;
+		}
+
+		private void Splitter_Shown(object o, EventArgs args)
+		{
+			if (!shown)
+			{
+				shown = true;
```
cwensley: You wouldn't need the shown bool if you just do `Shown -= Splitter_Shown;`

cwensley:
Also, what platform(s) have you tested this on?  I'd like to make sure this works on all platforms.  I don't mind doing some of the testing. (;

cra0zy:
@cwensley Internet ate my main comment so this was missing info :/ Anyway the main comment stated something along the lines:
 - I implemented this at friends house, it was done fast since we were working on something else (and only implemented for Horizontal orientation since we needed that)
 - I only tested on Gtk, I push this here so you can help me test on the rest
 - What do you want the name of the vars to be

Source/Eto/Forms/Controls/Splitter.cs
```diff
@@ -87,11 +89,40 @@ protected virtual void OnPositionChanged(EventArgs e)
 
 		#endregion
 
+		/// <summary>
+		/// Initializes a new instance of the <see cref="T:Eto.Forms.Splitter"/> class.
+		/// </summary>
+		public Splitter()
+		{
+			panel1MinimumSize = 0;
+			panel2MinimumSize = 0;
```
cwensley: Setting these to 0 is unnecessary, as that is the default value for int.

cra0zy: I know, it's just my preference to set the default values for all the variables in the constructor (so you can easily see what the starting value of something is).

cra0zy: Do you still want me to change this?

cwensley: Nah, it's all good.  This kind of pattern gets unwieldy when you have tons of members though, and it does actually generate slower code as it has to initialize the members twice.

cwensley: I'll do some testing on the other platforms before merging this in.  Thanks again for the PR!

cwensley:
Ok did some testing on this, seems to work well on macOS, but not on WinForms (Shown event actually doesn't work, but easily implemented), and has some problems on WPF as the position is set asynchronously so it jumps back and forth from where the mouse is and the minimum size set.

We may have to move the implementation of this to each of the handlers so they can implement it appropriately.

cra0zy:
@cwensley Gtk, WinForms and WPF implementations are done, I can't implement this for the Mac side (my VM is dead, and I don't have time to get it running again).

Issue #677, opened 2017-01-26
Changing the content of StackLayout does not respect VerticalContentAlignment


mennodeij:
When a control in a StackLayout is replaced with a new control, the stack layout's VerticalContentAlignment is not applied when the StackLayout is layed out again.
Repro: make a stacklayout (VerticalContentAlignment = Center) with a label, a combobox and a button. Upon button click, assign a new combo box to the Items[1] of the stacklayout. Result: the label goes from Center to Top vertical alignment.

mennodeij:
This happens on WPF and GTK.

Example:

    public class MainForm : Form
    {
        public MainForm()
        {
            Content = new StackLayout
            {
                VerticalContentAlignment = VerticalAlignment.Center,
                Orientation = Orientation.Horizontal,
                Items =
                {
                    new Label {Text = "Hallo"},
                    new ComboBox
                    {
                        Items =
                        {
                            "Aap",
                            "Noot",
                            "Mies"
                        },
                        SelectedIndex = 1
                    },
                    new Button
                    {
                        Text = "Click me!",
                        Command = ButtonCommand
                    }
                }
            };

        }

        private ICommand _buttonCommand;
        private ICommand ButtonCommand => _buttonCommand ?? (_buttonCommand = new RelayCommand<string>(OnCommandExecute));

        private void OnCommandExecute(string obj)
        {
            StackLayout s = Content as StackLayout;
            if (s == null) return;
            s.Items[1] = new ComboBox
            {
                Items =
                {
                    "Zus",
                    "Wim",
                    "Jet"
                },
                SelectedIndex = 1
            };
        }

cwensley:
Hey @mennodeij,

Thanks for the report and the repro!  Probably an oversight in the implementation.  I'll take a look!

Cheers,
Curtis.

cwensley:
Now fixed!  Thanks a bunch for the error report and repro.  I've turned your repro into a manual unit test as well.

PR #695, opened 2017-02-03
Fix label alignment in StackLayout when it is recreated.

cwensley:
- No longer update StackLayoutItem alignment after aligning the label so it remains the same value.
- Added manual test for StackLayout to reproduce the issue.
- Added TestBase.ManualForm with pass/fail buttons to make it easier to create more manual tests.
- Also added non-generic RelayCommand which I’ve been meaning to add for a while to make it easier to create a command that does not need a parameter.


cwensley:
This fixes issue #677

PR #710, opened 2017-02-14
DocumentControl and DocumentPage Gtk Implementations

cra0zy:
I'll work on a themed implementation of this in a separate PR.

Edit: Since I forgot to mention, a good chunk of code came from TabControl.

### DIFF
--- a/Source/Eto/Forms/Controls/DocumentControl.cs
+++ b/Source/Eto/Forms/Controls/DocumentControl.cs
@@ -1,0 +1,371 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Collections.ObjectModel;
+using System.Collections;
+
+namespace Eto.Forms
+{
+	/// <summary>
+	/// Arguments for the <see cref="DocumentControl"/> to get the current page.
+	/// </summary>
+	public class DocumentPageEventArgs : EventArgs
+	{
+		/// <summary>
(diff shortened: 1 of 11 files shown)


Source/Eto/Forms/Controls/DocumentControl.cs
```diff
@@ -0,0 +1,371 @@
...
+					this,
+					c => c.SelectedIndex,
+					(c, v) => c.SelectedIndex = v,
+					(c, h) => c.SelectedIndexChanged += h,
+					(c, h) => c.SelectedIndexChanged -= h
+				);
+			}
+		}
+
+		class DocumentPageCollection : IList<DocumentPage>
```
cwensley: Hm, any reason this isn't a `Collection<DocumentPage>`?

It seems that it might be better to use, as it'd eliminate the need for `IHandler.GetPage()`, as well as the need for most of the IList<T> implementation.

cra0zy: DocumentPages are reorderable :/

cra0zy: In case my comment wasn't clear, just think of how you can rearange tabs in web browser / programming IDE.

cwensley: Ah, yes, that makes sense.  Thanks for clarifying!

Issue #709, opened 2017-02-14
[Gtk] Label wrap issue


jaebird:
On Gtk2, when a particular refresh happens, a label that is configured to wrap, will wrap at about 50% of the label control. This can be reproduced by using the Test application under the Controls->Label section. Once visible, press the "Alt" key to cause the label to wrap. Resizing the window will reset the wrap correctly. In user applications, this happens at random times, however, this use case shows the same issue in a repeatable manner.

Tested on Ubuntu 16.10. I believe this issue has been present in Eto.Forms for awhile.

cra0zy:
Here is a sort of a workaround if you need it:
```c#
Style.Add<LabelHandler>("Wrap", h => h.Control.MaxWidthChars = 55);
```

And on the label just need to set the style:
```c#
label1.Style = "Wrap";
```

jaebird:
After a bit of digging, I've worked up a change to Eto.Forms that passes the tests I threw at it. This changes the Source/Eto.Gtk/Forms/Controls/LabelHandler.cs source code. Shown below in diff format:

> @@ -34,9 +34,20 @@ namespace Eto.GtkSharp.Forms.Controls
 			protected override void OnSizeRequested(ref Gtk.Requisition requisition)
 			{
 				int width, height;
\-				Layout.GetPixelSize(out width, out height);
\-				requisition.Width = width;
\-				requisition.Height = height;
\+				if (wrapWidth != null && wrapWidth > 0)
\+				{
\+					Layout.Width = (int)(wrapWidth * Pango.Scale.PangoScale);
\+					Layout.GetPixelSize(out width, out height);
\+					requisition.Width = width;
\+					requisition.Height = height;
\+				}
\+				else
\+				{
\+					Layout.GetPixelSize(out width, out height);
\+
\+					requisition.Width = width;
\+					requisition.Height = height;
\+				}
 			}
 			#else

I can work up a PR if required.

jaebird:
Another label issue was caused by the the one time OnRealized if a label was set as content to a form. The size of the label would not initially be set to the size of the form. The below diff fixes that issue:

>@@ -56,6 +67,25 @@ namespace Eto.GtkSharp.Forms.Controls
 				SetWrapWidth(allocation.Width);
 			}
\+ 
\+			protected override void OnRealized()
\+			{
\+				// the allocation may default to 1, in that case ignore OnRealized
\+				if (Allocation.Width > 1 && wrapWidth != Allocation.Width)
\+				{
[...]

PR #713, opened 2017-02-15
fixes Gtk2 label size and wrap issues

jaebird:
Fixes issue #709


cwensley:
Awesome!  Thanks for the fix.

Issue #715, opened 2017-02-17
[Gtk] RadioButton shows partial Label when Text is empty


jaebird:
If a radio button does not have text, the label should not be added to the control.

Tested on Gtk2.

PR #716, opened 2017-02-17
[Gtk] RadioButton shows partial Label when Text is empty #715

jaebird:
This is a partial fix, but need to ensure that RadioButtonHandler.Text is called when the Text of a RadioButton is changed.

### DIFF
--- a/Source/Eto.Gtk/Forms/Controls/RadioButtonHandler.cs
+++ b/Source/Eto.Gtk/Forms/Controls/RadioButtonHandler.cs
@@ -31,7 +31,8 @@ namespace Eto.GtkSharp.Forms.Controls
 				inactive.Active = true;
 			}
 			label = new Gtk.AccelLabel("");
-			Control.Add(label); //control.AddMnemonicLabel(label);
+			// don't add the label to the control unless Text is non-empty
+			//Control.Add(label); //control.AddMnemonicLabel(label);
 			Control.Toggled += Connector.HandleCheckedChanged;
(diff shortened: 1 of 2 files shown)


jaebird:
This fixes RadioButton with label Text. Should be ready to go. On a side note: RadioButtonList does not support changing Text once initialized. That does not affect this fix.

cwensley:
Thanks for the PR!  However, it appears there's a problem with the logic..  For example, if you run this code:

```c#
var rb = new RadioButton();

// problem scenario 1:
rb.Text = null; // will try to remove the label

// problem scenario 2:
rb.Text = "Something"; // adds the label, cool
rb.Text = "Something Else"; // will try to add the label again, not so cool
```

Cheers!
Curtis.

jaebird:
Is this blowing up in Gtk3? I tested in Gtk2 and it will not add or remove
a control unless it needs to. That is why I removed the flag logic. If
needed, I can put it back in.

cwensley:
Hm, to be honest I didn't actually test it, though are you sure GTK isn't spewing warning messages when this happens?  Ideally we should minimize those as much as possible.  If there aren't any warning messages, and this is perfectly legal GTK# code, then this should be fine.

jaebird:
I saw no warnings in Gtk2, which is why I removed the extra checks. I
haven't tested in Gtk3 though. I was assuming Gtk was tracking a dictionary
or something. :) I can test more.

cwensley:
Awesome!  Thanks for the PR.

PR #726, opened 2017-02-21
[Gtk] Size and ClientSize Fix

jaebird:
Builds on previous PR and fixes issue with the ClientSize and Size being consistent between both the Resizable and non-Resizable windows.


cra0zy:
@jaebird You shouldn't include my commit in your PR, also I rebased my PR so you have the wrong commit included as well...

jaebird:
Why don't you should just include this two line change. Your change fixes the resize when resizable is false. This change cannot be tested without it.

cra0zy:
> Why don't you just include this two line change. Your change fixes the resize when resizable is false. This change cannot be tested without it.

Your change is unrelated to what my PR is fixing.

PR #731, opened 2017-02-26
TrayIcon Implementation [Gtk, WPF, WinForms]

cra0zy:
Implements `TrayIndicator` and handlers for Gtk 2/3, WPF and WinForms.

There is one issue I found: <a href="https://social.msdn.microsoft.com/Forums/vstudio/en-US/7b169558-48b2-4bc9-9993-6327aca7cbe8/wpf-contextmenu-on-system-tray-icon-does-not-disappear-when-clicking-outside?forum=wpf">WPF Issue</a>, and the best possible solution for it would be <a href="https://weblogs.asp.net/marianor/a-wpf-wrapper-around-windows-form-notifyicon">this</a>, but I'll leave the implementation of that to someone else for now.


Source/Eto/Forms/TrayIndicator.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        /// <summary>
+        /// Sets the indicator menu.
+        /// </summary>
+        /// <remarks>
+        /// Make sure to call this event every time you make changes to the context menu.
+        /// </remarks>
+        /// <param name="menu">The indicator menu.</param>
+        public void SetMenu(ContextMenu menu)
```
cwensley: Could we change this to a property instead?  Could be done similarly to how you've done the Icon property.  It will also prevent the ContextMenu from being GC'd if the handler doesn't keep a reference to it.

Source/Eto.Mac/Eto.XamMac - net45.csproj
```diff
@@ -10,6 +10,7 @@
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <AssemblyName>Eto.XamMac</AssemblyName>
     <UseMSBuildEngine>True</UseMSBuildEngine>
+    <BaseIntermediateOutputPath>..\..\BuildOutput\obj\Eto.XamMac - net45</BaseIntermediateOutputPath>
```
cwensley: Not sure why these are being added to the csproj's, but they should be removed as they are set in Shared\Common.props

Source/Eto/Forms/TrayIndicator.cs
cra0zy: No, because stuff like libappindicator does not use a ContextMenu for example. What that method does is load the data from the ContextMenu.

Source/Eto.Mac/Eto.XamMac - net45.csproj
cra0zy: MonoDevelop decided to add them, dunno why...

Source/Eto/Forms/TrayIndicator.cs
cwensley: Could we not listen for changes to `ContextMenu.Items` and apply those as it is changed?

Source/Eto.Mac/Eto.XamMac - net45.csproj
cra0zy: Fixed, tho new line added at the end... some text editros follow the rule of adding a new line at the end while others don't, there is a whole debate on that...

Source/Eto/Forms/TrayIndicator.cs
cra0zy: Much easier said than done... plus this might not be the only backend that needs to load menu items so it's much easier to do this.

PR #732, opened 2017-02-27
Fsharp tutorials

smoothdeveloper:
* Converted C# tutorials
* add sample script to homepage

note: I'm using a file copy task in fsproj files instead of dependency on all eto platform assemblies.


cwensley:
This is really great!  Thanks for taking the time to put this together. 

It will make it much easier for F# developers to get started with Eto.Forms.

PR #733, opened 2017-02-27
[Gtk] fix Size and ClientSize for GtkWindow

k5jae:
This change fixes an issue when setting window sizes (resizable vs fixed). Using the Size value the result should include window decorations, when setting ClientSize, the window size should be the value plus window decorations.

Tested in Gtk2 and Gtk3


k5jae:
Test snippet:

	class MyForm : Form
	{
		public MyForm()
		{
			//this.WindowStyle = WindowStyle.None;
			Button button1 = new Button { Text = "r true" };
			button1.Click += (sender, e) =>
			{
				Resizable = true;
			};

			Button button2 = new Button { Text = "r false" };
			button2.Click += (sender, e) =>
			{
				Resizable = false;
			};

			Button button3 = new Button { Text = "client 400" };
			button3.Click += (sender, e) =>
			{
				ClientSize = new Size(400, 400);
			};

			Button button4 = new Button { Text = "size 400" };
			button4.Click += (sender, e) =>
			{
				Size = new Size(400, 400);
			};

			Content = new TableLayout
			{
				Spacing = new Size(5, 5), // space between each cell
				Padding = new Padding(10, 10, 10, 10), // space around the table's sides
				Rows =
				{
					new TableRow(
						new TableCell(button1, true),
						new TableCell(button2, true),
						button3,
						button4
					),
					new TableRow(
						new TextBox { Text = "Some text" },
						new DropDown { Items = { "Item 1", "Item 2", "Item 3" } },
						new CheckBox { Text = "A checkbox" }
					),
					// by default, the last row & column will get scaled. This adds a row at the end to take the extra space of the form.
					// otherwise, the above row will get scaled and stretch the TextBox/ComboBox/CheckBox to fill the remaining height.
					new TableRow { ScaleHeight = true }
				}
			};

		}
	}

cwensley:
@cra0zy let me know if this is good and I'll merge (;

harry-cpp (changes requested):

k5jae:
It does take an application tick from the set before the get returns the updated value, otherwise last size is retrieved.

Source/Eto.Gtk/Forms/GtkWindow.cs
```diff
@@ -246,13 +249,11 @@ public override Size ClientSize
...
 				{
-					var diff = vbox.Allocation.Size.ToEto() - containerBox.Allocation.Size.ToEto();
-					Control.Resize(value.Width + diff.Width, value.Height + diff.Height);
-					SetMinMax(value + diff);
+					Control.Resize(value.Width, value.Height);
+					SetMinMax(value);
 				}
 				else
 				{
-					clientSize = value;
```
cra0zy: Setting clientSize before control has allocated space wouldn't work so it has been set later on: https://github.com/k5jae/Eto/blob/322abd94a5e508e2e7f79c925e8d24ee9ee589b0/Source/Eto.Gtk/Forms/GtkWindow.cs#L266

With your changes Size and ClientSize would do the same thing before the control was realized.

harry-cpp (changes requested):

cwensley:
This is over a year old, should I close?  I don't think the issues with these changes were ever resolved.

cwensley:
Closing as the fix is probably stale and still has some issues as mentioned by @cra0zy.

PR #734, opened 2017-02-27
Fixes a crash when a bound property does not implement IConvertible.

MasterQ32:
This pull request improves the automatic conversion of property types when the bound property has another type than the bound value. It uses the [TypeConverter](https://msdn.microsoft.com/en-us/library/system.componentmodel.typeconverter(v=vs.110).aspx) system to find a fitting conversion implementation for the property type.


cwensley:
Looks like this was based off the master branch instead of develop.  Also, TypeConverter is not available in PCL 259 which Eto.dll uses so you'd need to use reflection to do the conversion instead.

cwensley:
Closing as this needs to be done in a way that works for .NET Standard, which doesn't have access to TypeDescriptor (this didn't work with PCL259 either).  Feel free to update the PR to use reflection to get at this functionality and ping me to reopen the PR and merge it.  I'm closing for now.

PR #735, opened 2017-02-27
Notification Implementation [Gtk, WPF, WinForms]

cra0zy:
Notes:
 - Gtk does not have a way of showing notifications on Windows and Mac
 - Someone else can add Windows 10 notification manager to WPF, for now it only has the standard way of using the StatusIcon to show notifications

### DIFF
--- a/Source/Eto/Forms/Notification.cs
+++ b/Source/Eto/Forms/Notification.cs
@@ -1,0 +1,172 @@
+﻿using System;
+using Eto.Drawing;
+
+namespace Eto.Forms
+{
+	/// <summary>
+	/// System toast notification.
+	/// </summary>
+	[Handler(typeof(Notification.IHandler))]
+	public class Notification : Widget
+	{
+		private Icon icon;
+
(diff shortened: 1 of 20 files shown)


Source/Eto/Forms/Notification.cs
```diff
@@ -0,0 +1,152 @@
...
+
+		/// <summary>
+		/// Shows the current notification.
+		/// </summary>
+		/// <remarks>
+		/// On some platforms like Gtk and Windows 10 the indicator
+		/// is not needed, while on the others like Windows 7 it's required.
+		/// </remarks>
+		/// <param name="indicator">Indicator to use to show the notification.</param>
+		public void Show(TrayIndicator indicator = null)
```
cwensley: Hm, how would one know they need a TrayIndicator here?  Perhaps we should add a `RequiresTrayIndicator` property of some sort so that you can create the TrayIndicator if it is required.

cra0zy: Done.

cwensley:
Awesome!!

PR #746, opened 2017-03-02
Mark TreeView as Obsolete

cra0zy:


Source/Eto/Forms/Controls/TreeView.cs
```diff
@@ -95,6 +95,7 @@ public TreeViewItemEditEventArgs(ITreeItem item, string label)
 	/// </remarks>
 	/// <copyright>(c) 2014 by Curtis Wensley</copyright>
 	/// <license type="BSD-3">See LICENSE for full terms</license>
+	[Obsolete("TreeView is deprecated, please use TreeGridView instead.")]
```
cwensley: Add "Since 2.3. " to the beginning of that.  Sorry, forgot to mention before. (;

cwensley: Err. 2.4

cra0zy:
@cwensley Because warnings as errors is enabled I have to disable some stuff with perma rules, I'll do it tomorrow.

cwensley:
I believe if you add `[Obsolete]` to the handler classes it would make it (fairly) easy.

cra0zy:
> I believe if you add [Obsolete] to the handler classes it would make it (fairly) easy.

You would than have to disable the rule in the Handler setup...

PR #751, opened 2017-03-03
Gtk: Optimize GraphicsPath to reduce memory & GC overhead.

cwensley:


cwensley:
@k5jae: would you mind running this through your profiling and let me know if there's anything else that's using excessive memory here?  I'm thinking I might have to make the Execute methods static, but don't know if it is allocating memory there.

k5jae:
Will do. Thanks!

cwensley:
@k5jae ok, my second take on optimizing GraphicsPath is now done.  I think this is probably as minimal as it'll get.  

GC pressure should be mostly minimal, there's only two lists - one for floats one for an enumeration.  The only objects created while adding to the path are the backing arrays in both the lists as they grow.  I've set the initial size of the lists to 16 commands and 64 floats.

k5jae:
@cwensley, testing shows good improvement. Is there a reason why there
isn't a Clear() method? I could probably use the same GraphicsPath object
in my render routine if there was. Either way, I think it is a good
improvement.

cwensley:
Thanks for testing that out, @k5jae!  

There's no `Clear` as no other platform allows for that.  Of course it would probably help with Gtk, but it wouldn't for other platforms.

PR #759, opened 2017-03-07
Fix so WPF controls have default style applied after the control is loaded

Linchy:
Previously some controls (such as combobox) would not get styled, as their visual tree was not yet loaded.

### DIFF
--- a/Source/Eto.Wpf/Forms/WpfFrameworkElement.cs
+++ b/Source/Eto.Wpf/Forms/WpfFrameworkElement.cs
@@ -532,8 +532,9 @@ namespace Eto.Wpf.Forms
 		}
 
 		void Control_Loaded(object sender, sw.RoutedEventArgs e)
-		{
-			SetSize();
+        {
+            SetDefaultStyleOnControlLoaded();
+            SetSize();
 			if (NeedsPixelSizeNotifications && Win32.PerMonitorDpiSupported)
 				OnLogicalPixelSizeChanged();
 		}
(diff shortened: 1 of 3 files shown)


Linchy:
Note: only tested on Windows, not sure if removing the style logic from Platform.cs and WidgetHandler.cs will affect other platforms now I come to think of it?

cwensley:
Hi @Linchy,

Thank you for the pull request!

Unfortunately, as you suspect, the changes would break style logic on all other platforms. On top of that, styling using the handler will also no longer work on WPF.

Also, there'd be no way to apply any changes to the control _before_ the control is loaded, which is a severe limitation.

Why would you need this?  If you want to create a style that performs some action on a WPF control after it is loaded, you can do something like this:

```c#
Style.Add<ComboBoxHandler>(null, handler => {
  handler.Loaded += (sender, e) => {
    // DoSomeStuff.
  };
});

```

Hopefully this helps!
Cheers,
Curtis.

Linchy:
Hi @cwensley yes I thought so. Well, thanks for the code then I would not of even thought of doing that. Perhaps add this to a section in the Styles guide? I was previously doing this: (which did not work at all on WPF):

```
Eto.Style.Add<DropDown>(null, s =>
{
    s.BackgroundColor = Colors.Black;
});

```

cwensley:
@Linchy that is indicative of a separate issue, it would be the same as doing:

`new DropDown { BackgroundColor = Colors.Black }`, which should work.

PR #762, opened 2017-03-13
[Gtk] Add init check to NotificationHandler

cra0zy:
I am just being paranoid O_o


cwensley:
Nothing wrong with being paranoid.. (;

PR #763, opened 2017-03-13
Implement taskbar progressbar

harry-cpp:
Notes:
 - Taskbar progressbar is implemented under Windows (WPF and WinForms backends) and under Ubuntu Unity and Ubuntu GNOME
   - ~~I have no idea if Mac has this, do tell me~~ It exists
 - If the taskbar handler is missing, the functions will do nothing instead of crashing, it made more sense
 - I flipped a coin between naming the main class `Taskbar` and `TaskbarButton`, do tell if you want it changed
 - ~~`AppId` property will be needed on Windows for showing notifications without tray icon so I put it in `Application` class~~ Not part of this PR, separate PR


cwensley:
Hm, is there any way to just generate an AppId?  Would a guid suffice, for example?

cwensley:
> There is not, in case of Linux the appid represents the app menu launcher file.

How would one know what to set this to? Would users have to set it differently depending on the platform?  What if someone created a linux version of their app, but not specifically for ubuntu for example, then someone created the .desktop file and ran it in ubuntu?

It's just likely that no one will ever set this properly, so if we can eliminate the need for it it would be ideal.

Source/Eto/Forms/Taskbar.cs
```diff
@@ -0,0 +1,38 @@
...
+		/// Sets the state and progress of the application.
+		/// </summary>
+		/// <param name="state">Taskbar button state.</param>
+		/// <param name="progress">Progress in range from 0.0f to 1.0f.</param>
+		public static void SetProgress(TaskbarProgressState state, float progress)
+		{
+			if (progress < 0.0f || progress > 1.0f)
+				throw new ArgumentOutOfRangeException(nameof(progress), "Progress needs to be in 0.0f to 1.0f range.");
+
+			var handler = Platform.Instance.Create<IHandler>();
```
cwensley: this should use `Platform.Instance.CreateShared<IHandler>()` so it only creates a single instance of the handler.

Source/Eto/Forms/Taskbar.cs
```diff
@@ -0,0 +1,38 @@
...
+	/// </summary>
+	[Handler(typeof(IHandler))]
+	public static class Taskbar
+	{
+		/// <summary>
+		/// Sets the state and progress of the application.
+		/// </summary>
+		/// <param name="state">Taskbar button state.</param>
+		/// <param name="progress">Progress in range from 0.0f to 1.0f.</param>
+		public static void SetProgress(TaskbarProgressState state, float progress)
```
cwensley: Should progress have a default of 0 here?  I assume it only makes sense to set its value when using TaskbarProgressState.ShowProgressbar

harry-cpp: Error and Pause states also use progress value, but the default value can't hurt.

cra0zy:
@cwensley What do you think on maybe requiring use of environmental variable on Ubuntu?

Source/Eto.Gtk/Platform.cs
```diff
@@ -174,11 +174,15 @@ public static void AddTo(Eto.Platform p)
 			p.Add<Keyboard.IHandler>(() => new KeyboardHandler());
 			p.Add<FixedMaskedTextProvider.IHandler>(() => new FixedMaskedTextProviderHandler());
 			if (EtoEnvironment.Platform.IsLinux)
+			{
 				p.Add<TrayIndicator.IHandler>(() => new LinuxTrayIndicatorHandler());
+				p.Add<Notification.IHandler>(() => new LinuxNotificationHandler());
+
+				if (Environment.GetEnvironmentVariable("DESKTOP_SESSION") == "ubuntu")
```
harry-cpp: @cwensley Just so you are not confused, this is not checking for Ubuntu, this is checking for Unity DE.

cwensley:
> What do you think on maybe requiring use of environmental variable on Unity?

Great idea - it puts the control of the name of the file into the file itself.

cra0zy:
Ok, the env variable is `DESKTOP_ENTRY`, if it's not set, the app will just output a message to the console that it needs to be set, this is so anyone running the app from terminal can easily spot it.

cwensley:
Awesome, that's great!

So I looked into macOS, and you can pretty much change the dock tile to contain any control(s).  Some apps do this to add a progress bar, which is certainly doable, e.g.

```c#
var app = NSApplication.SharedApplication;
var iv = new NSImageView();
iv.Image = app.ApplicationIconImage;

var pb = new ProgressBar();
pb.Value = 20;
var pbview = pb.ToNative(true);
pbview.AutoresizingMask = NSViewResizingMask.WidthSizable;
iv.AddSubview(pbview);
app.DockTile.ContentView = iv;
app.DockTile.Display();
```

The only thing that would be nice is to have a way to utilize macOS' bounce function, which can bounce the icon once (e.g. if something completed), or keep it bouncing if it is something requiring attention.  That could be done separately though.

cra0zy:
> The only thing that would be nice is to have a way to utilize macOS' bounce function, which can bounce the icon once (e.g. if something completed), or keep it bouncing if it is something requiring attention. That could be done separately though.

Do it like I did it for Ubuntu, on Ubuntu the function name is `unity_launcher_entry_set_urgent` and I call it if you set progress to 1f or if you set the state to error.

cra0zy:
> So I looked into macOS, and you can pretty much change the dock tile to contain any control(s). Some apps do this to add a progress bar, which is certainly doable, e.g.

You know it took me a bit of time to think about it, but that is so awesome.

cwensley:
> Do it like I did it for Ubuntu, on Ubuntu the function name is unity_launcher_entry_set_urgent and I call it if you set progress to 1f or if you set the state to Error.

That makes sense..  However, a lot of macOS apps use the bounce feature without any progress bar, so it'd be nice to allow for that.  Perhaps another method could be added to set alternate flags like urgent, notify, persistent notify, etc. ? 

I'm not asking you to do this right now, just trying to get ideas for how to address it in case it affects this api. (;

cra0zy:
Technically you can already do that, set state to `Default` and progress to `1f` :)

I would still add a separate method for just setting it, tho on Linux there is only urgent so I don't know what you mean with `notify` and `presistent notify`.

cra0zy:
Ubuntu Unity is... dead... yea...

cra0zy:
Since https://github.com/micheleg/dash-to-dock/pull/590 exists I removed restrictions for the taskbar progressbar API for Linux.

cra0zy:
> @cra0zy Now that unity is gone, is this still something that can be done on the Gtk side? Is there anything like this on newer gtk versions?

It works on GNOME with dash to dock extension (which is preinstalled on Ubuntu), so this is still a thing.

harry-cpp:
@cwensley I rebased this PR and improved it a bit, if you are interested in merging it.

harry-cpp:
PS. By default, works on newest versions of Ubuntu that come with GNOME, as the API and dash to dock are available on it.

harry-cpp:
PPS. Appveyor has internet problems XD

cwensley:
Fixed with #1660

PR #770, opened 2017-03-26
[GTK3] Avoid calling glue functions

cra0zy:


cra0zy:
@cwensley Can this be merged?

cwensley:
@cra0zy ah sorry I lost track of this one!  Are you sure there's no other use of the glue libs after this change?   

I'm thinking we should still keep them in the repo so the gtk# libs are complete, in the case where people are integrating with other apps, or writing gtk# code via styles.

Thoughts?

cra0zy:
> Are you sure there's no other use of the glue libs after this change?

Not 100% sure, but I couldn't find any other glue calls. I did test the Eto.Test application by removing my system gtk-sharp3 and it's glue libraries and it worked just fine.

> I'm thinking we should still keep them in the repo so the gtk# libs are complete, in the case where people are integrating with other apps, or writing gtk# code via styles.

If someone does call a glue function using styles, it will be called just fine since we are not providing gtk-sharp but instead using system gtk-sharp. (Side note, the glue libraries were useless since even the config files you added weren't using them :), also you only provided one set of them and you need 6 sets of them, 3 for each major system x 2 for x86 and x64 architectures )

cwensley:
> it will be called just fine since we are not providing gtk-sharp but instead using system gtk-sharp

Good point.  

On another side note, it would be nice to use your GTK# libs on nuget instead (assuming you're still maintaining it), but there's no way to stop it from copying the native libs over, which interferes with the GTK#2 test app.  It'd be nice if there was a condition put on the `CopyGtk3ConfigFiles` target so that the csproj can do something like `<EnableCopyGtk3ConfigFiles>false</EnableCopyGtk3ConfigFiles>` so they aren't copied.

cra0zy:
> On another side note, it would be nice to use your GTK# libs on nuget instead (assuming you're still maintaining it)

I am, but we can't use it since it's using a newer version of gtk-sharp :( Ie. you wouldn't be able to use system gtk-sharp because it's stuck on 3.0.

PR #776, opened 2017-04-05
Gtk3: Support child controls for Drawable.

cwensley:


cwensley:
@cra0zy could I get a review on this?  I'm pretty sure this won't have any problems, but want to make sure I didn't miss anything obvious.  I had to set the clip of the graphics context manually as it doesn't appear to be set up correctly when using `GLib.ConnectBefore`.

This is so you can add child controls to a Drawable.

cra0zy:
I don't think you missed anything.

cwensley:
Awesome!  I appreciate you double checking on that.

Issue #567, opened 2016-09-27
not support ilmerge?


ismlsmile:
I'm not sure if the problem is with eto.forms or ilmerge. When I use ilmerge to merge Eto*.dll into the exe file, the merge is ok, but when I run the output exe, it just crashes.

ismlsmile:
The exception is "Platform instance is null. Have you created your application?" in "Eto.Platform.get_Instance"

cwensley:
This is probably because when you are using auto-detection, it cannot load the appropriate DLL as it does it dynamically.  I'm not sure what you'll have to do to get around this, but most likely you'll have to create the particular platform class manually.. for example:

```c#
Platform platform;

if (EtoEnvironment.Platform.IsWindows) 
    platform = new Eto.Wpf.Platform();
else if (EtoEnvironment.Platform.IsMac)
    platform = new Eto.Mac.Platform();
else if (EtoEnvironment.Platform.IsLinux)
    platform = new Eto.Gtk.Platform();
else
    throw new Exception("Could not detect platform");

new Application(platform).Run(new MainForm());
```

I'm not sure if this will work 100% as it may have problems referencing each platform.

Hope this helps!
Curtis.

ismlsmile:
Thanks for your reply, I tried but it could not work, the exception changes to:

Current operation is valid only when the underlying type on a XamlType is known, but it is unknown for '{http://schema.picoe.ca/eto.forms}Label'

SlowLogicBoy:
I'm having the same problem, however I use only Eto.Wpf.

For this I use [il-repack](https://github.com/gluck/il-repack).
There are few problems:
1.  Eto.Wpf resources gets copied incorrectly
2.  Eto.Wpf resources are copied with name as for Eto.Wpf not with the new assembly name
3.  Eto.Wpf.Forms.ApplicationHandler.ApplyThemes uri is calling Eto.Wpf

I resolved this issue by using [dnlib](https://github.com/0xd4d/dnlib)
Here is the hack: [MyModulePatcher.csx](https://gist.github.com/SlowLogicBoy/ab7ececdbdbb8c2677f53bb0151a9550)

I hope this will help someone. There might be more issues, but for me these changes were enough.

PR #785, opened 2017-04-24
Removed hardcoded assembly name in Eto.Wpf #567

SlowLogicBoy:
#567

### DIFF
--- a/Source/Eto.Wpf/Forms/ApplicationHandler.cs
+++ b/Source/Eto.Wpf/Forms/ApplicationHandler.cs
@@ -77,7 +77,7 @@ namespace Eto.Wpf.Forms
 				return;
 
 			// Support (better) high DPI in the NumericUpDown control by theming the Extended WPF Toolkit's spinner.
-			var assemblyName = Assembly.GetCallingAssembly ().GetName ().Name;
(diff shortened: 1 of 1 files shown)


Source/Eto.Wpf/Forms/ApplicationHandler.cs
```diff
@@ -76,7 +77,8 @@ void ApplyThemes()
 				return;
 
 			// Support (better) high DPI in the NumericUpDown control by theming the Extended WPF Toolkit's spinner.
-			Control.Resources.MergedDictionaries.Add(new sw.ResourceDictionary { Source = new Uri("pack://application:,,,/Eto.Wpf;component/themes/wpftoolkit/ButtonSpinner.xaml", UriKind.RelativeOrAbsolute) });
+			var assemblyName = Assembly.GetCallingAssembly ().GetName ().Name;
```
cwensley: Might want to use `typeof(ApplicationHandler).Assembly` instead, this will use the assembly that calls this method so it is indeterminate.

SlowLogicBoy: Changed.

PR #793, opened 2017-05-01
Initial Commit of VisualStudioMac Addin Extension for Eto

anthonyirwin82:
I used the existing Xamarin Studio Extension and updated it to compile in the Visual Studio Mac preview edition with all the updates installed.

It compiles against the mono 5.x runtime fine.

I have the compiled dll files but not sure how to package it into a .mpack file to install into the editor.


cra0zy:
There is literally no point in having both XS and VS Mac extensions... you either need one or the other. Right now XS extensions is needed, but once VS Mac gets released the other will be the needed one. Also you can achive the support for both with few `#if`s and 2 project files without copying over the entire extension folder.

cwensley:
Hey @anthonyirwin82, this is really great!  Thanks for putting in the effort to get this up to date. 

I do agree with @cra0zy though in that this should probably just be another csproj with `#if`'s to make the code compile on both MD and XS.  Did you have to make a ton of changes, or was it pretty straight forward to update to VS on Mac?

Thanks again!

anthonyirwin82:
Ok yeah that makes sense. I just wanted to get it working with VS for Mac without screwing any of your stuff up.

To make this version work properly without any exception errors being thrown when loading any project in VS for Mac I had to Compile against .net 4.6.2 not 4.5

the mdtool is also now called vstool

To get it to work I had to replace all the 2.4.0-alpha references to 2.3.0 as the 2.4.0-alpha doesn't have any packages to download.

I have built a .mpack file and tested with the default eto.forms app in c# format and it works and compiles fine.

If Visual Studio for Mac users want it they can download it here:
https://github.com/anthonyirwin82/Eto/blob/develop/Visual-Studio-Mac-Extension-For-Eto.Forms-in%20mpack%20format/Eto.Addin.VisualStudioMac_2.3.0.6.mpack

I know you guys want to do the changes in a proper steamed line way for 2.4 release of eto so I guess anyone who wants to use VS for mac with 2.3.0 can use this third party install file I created.

Hope it helps out end users.

I am sure you guys can get it to work with both XS and VS for mac with conditional if statements but I am not proceeding any further with it I just wanted it working in VS for Mac and this quick hack did the trick for me.

Sappharad:
FYI, the provided build doesn't play nice with the dark theme. Probably pre-existing though.
screen shot 2017-06-27 at 7 41 22 pm

cwensley:
Unfortunately, since you're not planning to update this any further with this we'll have to do it differently.  Thanks for submitting the PR and making this available to other users!  It has been a great stop-gap until it's fully baked.

I'm going to be updating the addins to use [dotnet templates](https://github.com/dotnet/templating) instead, so we can simplify things and use the same templates for all platforms.

Issue #752, opened 2017-03-05
[WPF] Unable to set Text property in ComboBox


master131:
```c#
var cb = new ComboBox { Text = "ABC" };
```

The above code (in InitializeComponent) leads to the following exception being produced:
```
An unhandled exception of type 'System.ArgumentNullException' occurred in System.Core.dll

Additional information: Value cannot be null.

   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at Eto.Wpf.Forms.Controls.ComboBoxHandler.HandleTextChanged()
   at Eto.Wpf.Forms.Controls.ComboBoxHandler.set_Text(String value)
   at Eto.Forms.ComboBox.set_Text(String value)
   at SampleApp.MainForm.InitializeComponent() 
```

First time using Eto so I'm unsure whether this is actually a bug or not. I used the VS Extension Template with the PCL + Code Preview options. That line would be valid in a standard WPF project given that the IsEditable property was true to show the custom-set text; it would not throw an exception even if it were set to false though.

EDIT - After snooping around in the source, the following appears to stop the exception from appearing:
```c#
var cb = new ComboBox { DataStore = new object[] {}, Text = "ABC" };
```

PR #812, opened 2017-05-26
Wpf: Fix regression when controls are set to their default size.

cwensley:
Wpf: Fix setting text before setting data store
Wpf: ProgressBar and Slider should have sane default sizes


cwensley:
Should fix #752

Issue #789, opened 2017-04-24
Wpf: a DrawableCell in a TreeGridView is rendered all white if it is in the first column 


msasso69:
A DrawableCell in a TreeGridView if it is the first column is always rendered all white (see below picture)

![image](https://cloud.githubusercontent.com/assets/21080030/25361568/3d576edc-294f-11e7-979c-8c8e7d0bbf97.png)

The following code reproduces the problem:

    using System;
    using Eto;
    using Eto.Forms;
    using Eto.Drawing;

    namespace VisualSEO.EtoFormGui
    {
        public class Program
        {
            [STAThread]
            public static void Main(string[] args)
            {
                new Application(Platform.Detect).Run(new MyForm3());
            }

            public class MyForm3 : Form
            {
                public MyForm3()
                {
                    TreeGridView tree = new TreeGridView();

                    // add first drawable (it will be rendered all white!
                    var drawableCell1 = new DrawableCell();
                    drawableCell1.Paint += drawableCell_Paint;
                    tree.Columns.Add(new Eto.Forms.GridColumn { HeaderText = "not working", DataCell = drawableCell1, AutoSize = true });

                    // add the second drawable
                    var drawableCell2 = new DrawableCell();
                    drawableCell2.Paint += drawableCell_Paint;
                    tree.Columns.Add(new Eto.Forms.GridColumn { HeaderText = "working", DataCell = drawableCell2, AutoSize = true });

                    // add some data to the tree just to populate it
                    TreeGridItemCollection model = new TreeGridItemCollection();
                    TreeGridItem item1 = new TreeGridItem();
                    item1.Values = new object[] { "text" };
                    TreeGridItem item2 = new TreeGridItem();
                    item2.Values = new object[] { "text" };
                    item1.Children.Add(item2);
                    item1.Expanded = true;
                    model.Add(item1);

                    tree.DataStore = model;

                    Content = new StackLayout
                    {
                        HorizontalContentAlignment = HorizontalAlignment.Stretch,
                        Items =
                        {
                            tree
                        }
                    };
                }

                void drawableCell_Paint(object sender, CellPaintEventArgs e)
                {
                    var m = e.Item as TreeGridItem;
                    if (m != null)
                    {
                        string text = (string)m.Values[0];
                        e.Graphics.DrawText(SystemFonts.Label(), Colors.Black, (float)0, (float)0, text != null ? text : string.Empty);
                    }
                }
            }
        }
    }

msasso69:
[...]

PR #821, opened 2017-06-08
Wpf: fix first column all white DrawableCell in TreeGridView (#789)

msasso69:
A DrawableCell in a TreeGridView (if it was the first column) was always rendered all white,
the RenderSize width needs to be set explicitly based from the column width.

### DIFF
--- a/Source/Eto.Wpf/Forms/Cells/DrawableCellHandler.cs
+++ b/Source/Eto.Wpf/Forms/Cells/DrawableCellHandler.cs
@@ -26,6 +26,7 @@ namespace Eto.Wpf.Forms.Cells
 			protected override void OnRender(swm.DrawingContext dc)
 			{
 				var handler = Column.Handler;
+				RenderSize = new System.Windows.Size(this.Column.ActualWidth, RenderSize.Height);
 				var graphics = new Graphics(new GraphicsHandler(this, dc, new sw.Rect(RenderSize), new RectangleF(RenderSize.ToEto()), false));
 				var state = IsSelected ? CellStates.Selected : CellStates.None;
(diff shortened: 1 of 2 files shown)


msasso69:
I now notice that in my pull request went two fixes, I would have liked to make two separate pull requests (one per fix), but it is my first pull request and I'm quite new with git. I'll do better in my next fixes ;)

cwensley:
Hey @msasso69 thanks for the PR!

For your fix with a hyperlink, it should be more generically to handle other non-visual elements.  You can traverse the parent tree [this way](https://github.com/picoe/Eto/commit/3e14e334b7d92af8edb7267eee43468fd7495230#diff-ffb0b14dce33fd1efacf61f2419e0d4cR134) instead, which should make sure the HasFocus actually works.

Perhaps we should create a GetVisualParents() extension so that it can deal with this in a single place?

msasso69:
I made it now "generic" doing so:

image

now it makes much more sense ;)
of course it has to be added also the following line at the beninning:

image

Surely an extension as you suggest can also be done if we need to reuse the same code elsewhere, but do we need to do the same also somewhere else? I leave it to your infinite wisdom! ;)

cwensley:
Thanks again for the fix, but I found that with what you proposed a) WPF's documentation says [not to modify RenderSize directly](https://msdn.microsoft.com/en-us/library/system.windows.uielement.rendersize(v=vs.110).aspx), and b) the ClipRectangle would extend beyond the size of the actual cell.

I've fixed it by using a DockPanel instead of a StackPanel for the expander and content so that the content of the cell will stretch to fill regardless of its desired size.

PR #863, opened 2017-08-17
Make Wpf DialogHandler more extendable

SlowLogicBoy:
I don't know if this will work for everyone, but this works for me


cwensley:
Thanks for the PR!  I think it'd be nice to have a way to construct it without necessarily initializing all of its properties to integrate better with native apps, but that can be done when needed.  (;

PR #880, opened 2017-09-06
In progress: Drag 'n Drop for the masses.

cwensley:


SlowLogicBoy:
I need to try this like right now.

cwensley:
@SlowLogicBoy Sorry, you probably won't get very far as it's a work in progress. It doesn't really work on anything other than WPF at the moment, and it has issues to boot.

cwensley:
I'm currently starting to implement this for macOS, which will by far be the hardest part of this as its drag 'n drop infrastructure is quite different from windows or gtk.

SlowLogicBoy:
Well, I'm using Wpf platform mainly, I just want to see how mock up looks like :) and I bet this will change with time, just too excited about this :)

cwensley:
Ok, I can guarantee there's going to be bugs with this one.  I could use some help testing the different platforms and features. Anyone have some free cycles for testing?

TomQv:
Cool ! Long awaited, Thanks.
Will try Wpf and Mac the next days.
Tom

PR #891, opened 2017-09-27
Gesture support - Issue 413

derbyw:
The following is the start of addressing Issue 413 which adds gestures (mouse and multi-touch). At the moment only GTK#3 has been added. The code was built against gtksharp built against Gtk+-3 3.22.8 so there are also some mods here dealing with gtk functions marked obsolete

Eto:
     to Eto. In Eto.Forms.Control it adds gesture events and modifies Control to add support.

Eto.Gtk:
     for GTK# 3 wire the gesture events to the Eto abstractions. 

GesturesEto and GesturesEto.Gtk3
      A simple Eto test app that works with gestures.


cwensley:
Hey @derbyw, thanks for the PR!  This would be excellent to get in.

It unfortunately looks like you branched from **master**.  You'll need to rebase your changes on the **develop** branch instead and re-sync the branch.

derbyw:
Hi Curtis,


   OK - sorry - I wasn't sure about which to use.


William Derby
derbyw@verizon.net




-----Original Message-----

cra0zy:
Also gtk-sharp 3 must not get updated, you'll have to write the wrappers for the functions that are missing from gtk-sharp 3.

derbyw:
> Also gtk-sharp 3 must not get updated, you'll have to write the wrappers for the functions that are missing from gtk-sharp 3.

Hmm - tricky since the "stock" gtk-sharp3 was missing the gtk gesture events and definitions and was one of the reasons I built a gtk-sharp3 based on the at the time current rev (3.22.8). Will gtk-sharp3 be stuck at this rev forever for Eto? it's getting pretty old and GTK4 will be out soon. I notice that you (Cra0zy) are doing a native GTK port for Eto, and also have a updated gtk-sharp nuget package -- what gtk rev is the native one targetng?

May I propose that the the pull request really has 2 parts - the first adds gesture supports to Eto itself and is pure and could be merged now. The second which is the GTK3 specific implementation has the version/update issues issues w/ gtk-sharp3 . i.e. the commits to the Eto - pcl tree would allow Eto apps to register for gesture events and when/as the various native implementations are updated, the Eto apps gesture handling would "just work". 

Failing that can you point me to an example of a GTK wrapper - I don't really want to reinvent the wheel on this...

cra0zy:
> Will gtk-sharp3 be stuck at this rev forever for Eto?

Once NativeGtk backed is done, the idea was to delete Gtk 3 backend.

> what gtk rev is the native one targetng?

Newest possible Gtk version, with flatpak out its not so problematic targeting it.

cwensley:
Closing this as #892 replaces this PR.  Please move discussion there.

PR #892, opened 2017-09-28
Gestures 413develop

derbyw:
Same as last one but cleaned up and re-targeted to the develop branch. As mentioned before, this branch is targeted to GTK+-3 v. 3.22.8 which I needed to target for my applications. Correspondingly I I have a gtk-sharp3.3.22.8.0.nupkg developed that I can send as I'm not sure how to include with this project. If you look at the commits I simple used the package manager to target the nuget package (which replaced all the old gtksharp libraries). let me know how best to get you what you need to evaluate it. If need be I guess I can create a git repo for it. 

As before this includes 2 test projects one of which is an Eto project and the other is the GTK variant. I don't think I can do a GTK2 version but I can probably eventually do a windows/wpf implementation as I get time or have need.

In the solution, "warnings as errors" is also turned off due to a large amount (85+) of obsolete GTK calls still in the target. I'll try to work through these as time permits but each one is a mini research project.


derbyw:
As noted in PR 891 comments - due to conflicts with gtk-sharp3, it may be best to just incorporate the changes to Eto - pcl which will add gesture events to Eto controls and then deal with separate native libraries as stand-alone PR's. Can you pull that out of this PR or do I need to submit a 3rd PR?

cwensley:
Hey @derbyw, thanks for rebasing on develop, it makes it much easier to see the changes you've made.   I really appreciate the time and effort you put into this!  Here's some feedback that will hopefully be useful:

1. We unfortunately cannot update the gtk-sharp3 binaries as it would break usage on platforms such as ubuntu.  The Gtk3 platform must remain compatible with the readily available gtk-sharp3 binaries on most/popular linux distros.  Also, targeting the latest version will effectively make the app break if it is running on any version of gtk+ lower than what you target (afaik).
2. To implement features that gtk-sharp doesn't expose, you can use native calls, much like how [`ColorDialogHandlerGtk34`](https://github.com/picoe/Eto/blob/c136c6f2014c32c4d45b77cf0dfd8c915a0e432c/Source/Eto.Gtk/Forms/ColorDialogHandlerGtk34.cs) is implemented. This will also be more easily transferrable to the NativeGtk port that @cra0zy is working on that will eliminate the need for the gtk-sharp dependency altogether.
3. I am currently working towards finishing up Eto 2.4, so I won't be able to pull this unless it is complete on the major platforms (WinForms, WPF, Mac/XamMac, Gtk).  
4. After 2.4 is released I can pull.  Pulling just the API into Eto - pcl is not really useful if no platforms will implement it.  The API should be flushed out by having a few platform implementations first, as the API design often needs tweaking to support all platforms properly. 
5. There needs to be a way to determine if a platform supports gestures at runtime.  This can be done by adding to the [PlatformFeatures](https://github.com/picoe/Eto/blob/develop/Source/Eto/Platform.cs#L70) flags.
6. Lastly, please incorporate the new tests into Eto.Test instead.  Having multiple test apps for each feature is not very maintainable and would be bit rotted very quickly as I cannot maintain them.

Hope all this feedback helps!

Cheers,
Curtis.

cwensley:
As for the API/events you added, I do like how it's organized.  The only thing I think might make sense to change would be to combine the PanH and PanV events into one (though this would need investigation first).  I'm not sure what Gtk3 supports but on macOS for example you can pan both directions at the same time which would be difficult to use with two events.

cwensley:
Hi @derbyw,

Thanks a ton for your PR.  Now that we are using the latest Gtk# from @cra0zy, this should be updated to use that instead, and also address the comments.  E.g. we will also need implementations for windows and mac before this can be pulled.

I'm closing this, but please feel free to update with the latest code and provide implementations for the other platforms and address the other comments.

Issue #862, opened 2017-08-16
Migrate PCL to .NET Standard


cwensley:
Currently Eto.dll, Eto.Serialization.Xaml, and Eto.Serialization.Json all use PCL 259.

They should be migrated to .NET Standard, hopefully in the 1.x range to support the largest number of platforms.

All project templates should also be updated to create .NET Standard projects instead of PCL for the shared UI logic assembly.

cwensley:
Now done by PR #949

PR #914, opened 2017-11-10
Make Eto - pcl a .NET Standard library

cra0zy:
Fixes: #862


cwensley:
Now in!  thanks for the initial work on this.

PR #949, opened 2018-01-15
Add .NET Standard support

cwensley:
This PR changes a _few_ things:

- Eto.dll is now an .NET Standard 1.0 assembly (compatible with pretty much everything)
- Eto.Serialization.Json/Xaml now provide an .NET Standard 1.3 assembly as well as Portable259.
- Folder structure now follows [this](https://gist.github.com/davidfowl/ed7564297c61fe9ab814)
- Remove using built-in nuget.exe for restores during build (still used for CI build)
- Migrate some projects to the sdk-style csproj format
- Deprecated Eto.Platform.Mac.Template as it is now rolled into Eto.Platform.Mac/Mac64.
- Eto.Platform.Mac/Mac64 nuget packages now enable direct run/debug from VS for Mac


src/Eto/Eto.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard1.0</TargetFramework>
+	<GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DefineConstants>TRACE;DEBUG;PCL</DefineConstants>
+    <OutputPath>..\..\artifacts\core\Debug\</OutputPath>
```
SlowLogicBoy: You can move this to main `<PropertyGroup>`
`<OutputPath>..\..\artifacts\core\$(Configuration)\</OutputPath>`

src/Eto/Eto.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard1.0</TargetFramework>
+	<GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DefineConstants>TRACE;DEBUG;PCL</DefineConstants>
```
SlowLogicBoy: `<DefineConstants>TRACE;PCL</DefineConstants>`
I think you can move this to main `<PropertyGroup>` since Debug const is added by default on Debug build (if newly defined ones doesn't override.
And by moving both (DefineConstants and OutputPath) you could get rid of these PropertyGroups with conditions

src/Eto/Eto.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <OutputPath>..\..\artifacts\core\Debug\</OutputPath>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DefineConstants>TRACE;PCL</DefineConstants>
+    <OutputPath>..\..\artifacts\core\Release\</OutputPath>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="..\Shared\EnumerableExtensions.cs" />
+    <Compile Include="..\Shared\GlobalAssemblyInfo.cs" />
```
SlowLogicBoy: You can ether create a Eto.props file with stuff that's in AssemblyInfo.cs and import them, or you can add AssemblyInfo properties to csproj itself example:
```xml
    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
        <DebugType>Full</DebugType>
        <Title>MyApp</Title>
        <Authors>MyCompany</Authors>
        <Description>My Description</Description>
        <Copyright>My Copyright</Copyright>
        <Version>1.0.0.0</Version>
    </PropertyGroup>
```
These are also added into nuget package.
For globals we do:
MyProduct.targets:
```xml
<Project>
    <PropertyGroup>
        <Authors>My Company</Authors>
        <Owners>$(Authors)</Owners>
        <Company>$(Authors)</Company>
        <CopyrightYear>2018</CopyrightYear>
        <Copyright>(c) 2010-$(CopyrightYear) by Curtis Wensley, 2012-2014 by Vivek Jhaveri and contributors</Copyright>
        <Product>My Product</Product>
        <PackageIconUrl>url to nuget icon</PackageIconUrl>
        <RepositoryUrl>Path to repository</RepositoryUrl>
        <RepositoryType>Git</RepositoryType>
        <GenerateDocumentationFile>true</GenerateDocumentationFile>
        <DocumentationFile>$(OutputPath)$(AssemblyName).xml</DocumentationFile>
    </PropertyGroup>
</Project>
```
MyProject.csproj:
`<Import Project="../path/to/MyProduct.targets" />`

src/Eto/Eto.csproj
cwensley: Hm, I'll try that, thanks!

src/Eto/Eto.csproj
cwensley: I tried putting this outside all projects in the shared props file, but I think it's because I was trying to be too aggressive.  I think you're right in that for sdk-style csproj's I can do this.

src/Eto/Eto.csproj
cwensley: Yeah I know about this, but since I need the GlobalAssemblyInfo for all the other non-sdk style projects I won't be able to do this afaik.  Also, without multi-targeting support in VS for Mac I won't be able to use it for the xaml/json assemblies either so I will keep the nuspec files for now.  I may transition to this later when the tooling catches up..

cra0zy: Yea, multi targeting support is not available on Linux either, tho this is not a bug in VS for Mac/MD, its a bug with msbuild. There are 2 msbuilds that get installed and neither of them are able to see the framework the other one supports:
 - `msbuild` - only sees standard .NET Framewrok
 - `dotnet msbuild` - only sees .NET Core frameworks

On a side note, they are also linked against only the framework they support so you actually have to build 2 C# task files if you want to do custom tasks.

PR #953, opened 2018-01-19
.NET Core Gtk backend

cra0zy:
I probably need to cleanup something, I'll do it tomorrow, good night :)


src/Eto/TypeHelper.cs
```diff
@@ -14,9 +14,15 @@ namespace Eto
...
-
-		public static MethodInfo GetCallingAssembly { get { return getCallingAssembly; } }
+        public static Assembly GetCallingAssembly()
+        {
+            var getCallingAssembly = typeof(Assembly).GetTypeInfo().GetDeclaredMethod("GetCallingAssembly");
+            var detectType = Type.GetType("Eto.PlatformDetect, Eto.Gtk", false);
+            if (detectType != null)
+                getCallingAssembly = detectType.GetRuntimeMethod("GetCallingAssembly", new Type[] { });
+
+            return getCallingAssembly.Invoke(null, null) as Assembly;
```
cwensley: afaik, this will always return Eto.dll, since it's being called from Eto code..  hence is why the .Invoke was repeated everywhere before.

test/Eto.Test/Eto.Test.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
cwensley: I pondered this, but decided to keep Eto.Test as net standard 1.6.  It reduces the required dependencies so you don't have to install .net core sdk to compile everything.  That being said, the new gtk platform will require it so maybe that's just better.  Just a brain dump, let me know if you feel strongly either way.

cwensley:
I'm super excited about this, thanks for all the hard work!!  Looks like there's a compile error in Gtk2 platform after these changes, should be easy to fix.

BTW, what IDE are you using?  I think i'll have to re-add the MD/VS for Mac stuff to Eto.csproj to get the formatting correct (tabs), assuming that's what you're using.

src/Eto.Gtk/Eto.Gtk.csproj
```diff
@@ -0,0 +1,47 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="GtkSharp" Version="3.22.24.28" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Eto\Eto.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Remove="EtoWebView.cs" />
+    <Compile Remove="Forms\Controls\WebViewHandler.cs" />
```
cwensley: Ah, boo.  Any idea if this can be done in the future?  A lot of projects use WebView.

src/Eto/TypeHelper.cs
cra0zy: You have not looked at the 2nd part of code, its not invoking an actual GetCallingAssembly, but instead a StackTrace.

test/Eto.Test/Eto.Test.csproj
cra0zy: This was a mistake on my part XD

src/Eto.Gtk/Eto.Gtk.csproj
cra0zy: That's the Gtk 2 webview, there is separate Gtk 3 webview.

cra0zy:
> BTW, what IDE are you using? I think i'll have to re-add the MD/VS for Mac stuff to Eto.csproj to get the formatting correct (tabs), assuming that's what you're using.

Yea, I used MD.

src/Eto/TypeHelper.cs
cra0zy: Oh, I understand what you meant now, will fix it ASAP.

cwensley:
Excellent!!

PR #970, opened 2018-01-26
Remove all PCL!

cwensley:
- Eto.Serialization.Json and Xaml are now both .NET Standard 1.0
- Include .targets for json/xaml to embed the necessary files (when using sdk style csproj's) and set up DependentUpon
- Only generate mac bundle if not using a specific xamarin.mac or monomac project type


cra0zy:
RIP (Rest in PCL).

PR #976, opened 2018-02-01
Few fixes for .NET Core & Gtk

cwensley:
- Fix platform detection on .NET Core
- Fix getting (proper) calling assembly on Gtk platform
- Set up library fallback path so we can run Eto.Test.Gtk on macOS

### DIFF
--- a/src/Shared/PlatformDetect.cs
+++ b/src/Shared/PlatformDetect.cs
@@ -23,7 +23,7 @@ namespace Eto
 			{
 				buf = Marshal.AllocHGlobal(8192);
 				if (uname(buf) == 0)
-					osName = Marshal.PtrToStringAuto(buf);
+					osName = Marshal.PtrToStringAnsi(buf);
 			}
 			// Analysis disable once EmptyGeneralCatchClause
 			catch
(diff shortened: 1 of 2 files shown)


src/Shared/PlatformDetect.cs
```diff
@@ -40,7 +40,7 @@ public static string GetUnixType()
 #if GTK3
 		public static Assembly GetCallingAssembly()
 		{
-			return (new StackTrace()).GetFrame(5).GetMethod().DeclaringType.Assembly;
+			return (new StackTrace()).GetFrame(4).GetMethod().DeclaringType.Assembly;
```
cra0zy: @cwensley I think you need to check the frame name, it seems like the index is not constant between platforms?

cwensley: Are you sure?  Your initial implementation was actually calling this method within a method before, but now that it's called using `.Invoke()` it's one less frame than it was initially.  At least that is my theory (and observation with the `AboutDialog`).  I thought of traversing the stack till it got to Eto, then until it wasn't Eto and return that, but thought it'd be pretty slow.

cra0zy: Hmm, possibly, for now it does work so I guess we'll see if someone complains.

cra0zy: @cwensley I was right, we can't use a static number, or at least the number is different between mono and .net core. (I don't have the time to fix it right now)

Issue #938, opened 2017-12-22
Memory report on Eto.Test does not show process memory (hiding un-managed memory leaks.)


derbyw:
The current "Memory" display on the Eto.Test app only shows only managed memory usage. If there are memory leaks in un-managed memory owned by Eto (i.e. handles, contexts, etc.) they will not be apparent in the test app.  I propose that in addition to the GC.GetTotalMemory(true)); that we also display
the un-managed memory owned by the process i.e. 

using System.Diagnostics;

......
in MemoryButton handler:

using (Process ThisProcess = Process.GetCurrentProcess()) {
       Log.Write(null, "GC Memory: {0}", GC.GetTotalMemory(true));
       Log.Write (null, "Process Total {1}", ThisProcess.PrivateMemorySize64);
}

derbyw:
Sigh - This will need to wait for when/if the code is switched to .NET standard - the pcl System dll apparently doesn't include a System.Diagnostics.Process..

cwensley:
Now in!  Thanks for the suggestion.

PR #979, opened 2018-02-01
Show native memory usage in test app

cwensley:


cwensley:
Wow, fun to edit directly in github.. slow to do a recompile, but it works!

cwensley:
Yes, I'm talking to myself..

cwensley:
Fixes #938

PR #996, opened 2018-02-09
Only restrict to 32-bit with Gtk2 on windows

cwensley:


sourceskyboxer:
Please use Gtk 3 It used from msys64\mingw64\bin = Gtk runtime 3.x 64 bit
and msys64\mingw32\bin = Gtk runtime 3.x 32 bit Please fix and check like https://github.com/robertsundstrom/gtk-core Thanks!

cwensley:
@sourceskyboxer thanks for the comment, but I'm not sure what you're asking.. I'm only removing the 32-bit restriction for GTK3 so it can run in 64-bit.

sourceskyboxer:
@cwensley Welcome but it can work x86 and x64 bit too because msys64 has 2 i686 and x86_64 gitk runtime check website http://grbd.github.io/posts/2016/01/29/gtksharp-part-2-setting-up-glade/

cwensley:
Yes, that is what this change has done.. it is making it so only GTK2 is restricted to 32-bit.  GTK3 will run on both 32 and 64 bit on windows.

sourceskyboxer:
And do not forget to add `Control.TypeHint = Gdk.WindowTypeHint.Desktop;` in FormHandler in Eto.Gtk3
Because it forgets to use maximize/restore 

See my issue #1001 

Thanks!

PR #1005, opened 2018-02-13
[Gtk] Fix setting messagebox text

cra0zy:
The constructor for the messagedialog does not take in a text param, but instead a string format. In case you were to use `{ }` in any way this would cause a crash.

### DIFF
--- a/src/Eto.Gtk/Forms/MessageBoxHandler.cs
+++ b/src/Eto.Gtk/Forms/MessageBoxHandler.cs
@@ -23,7 +23,8 @@ namespace Eto.GtkSharp.Forms
 			if (parent != null && parent.ParentWindow != null)
 				parentWindow = parent.ParentWindow.ControlObject as Gtk.Window;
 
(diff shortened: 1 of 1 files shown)


src/Eto.Gtk/Forms/MessageBoxHandler.cs
```diff
@@ -23,7 +23,8 @@ public DialogResult ShowDialog(Control parent)
 			if (parent != null && parent.ParentWindow != null)
 				parentWindow = parent.ParentWindow.ControlObject as Gtk.Window;
 
-			control = new Gtk.MessageDialog(parentWindow, Gtk.DialogFlags.Modal, Type.ToGtk(), Buttons.ToGtk(), false, Text);
+			control = new Gtk.MessageDialog(parentWindow, Gtk.DialogFlags.Modal, Type.ToGtk(), Buttons.ToGtk(), false, "");
```
SlowLogicBoy: how about `string.Empty` ?

PR #1007, opened 2018-02-13
Fix binding null SelectedPage exception

SlowLogicBoy:


cwensley:
Hey thanks for the PR, @SlowLogicBoy!

I think we should set the SelectedIndex to -1 if null is passed, instead of doing nothing.

SlowLogicBoy:
Yea, I am afraid of doing that, because it might break something else? like plarforms?

cwensley:
Sweet, thanks for the update.  If it does crash when setting it to -1 in that case then it'd be a bug further down the chain.

PR #1018, opened 2018-02-17
[GTK] Fix GetCallingAssembly

cra0zy:


cra0zy:
@cwensley PS. The error with this is between .NET Core and .NET Framework / mono, they do have a different amount of calls in the stacktrace.

cwensley:
Hm, a few thoughts.. 

1. Perhaps we should be comparing the Assembly directly, not just the name?  
2. Also, it might make sense to find Eto, then continue until the assembly that is not Eto.. unless that is where the extra frame(s) are located..

cra0zy:
> Perhaps we should be comparing the Assembly directly, not just the name?

I thought that as well at the start, but then I realized that we really only care about those few names, and they are gonna be unique anyways.

> Also, it might make sense to find Eto, then continue until the assembly that is not Eto.. unless that is where the extra frame(s) are located..

Can't do that because the frames go something like: Eto.Gtk -> mscore/System -> Eto -> Calling Assembly

cwensley:
> but then I realized that we really only care about those few names, and they are gonna be unique anyways.

True, though I was thinking more for performance, loading a ton of bitmaps/icons for example.

> Eto.Gtk -> mscore/System -> Eto -> Calling Assembly

That's what I mean..  Find Eto, then find the next assembly after it.  No need to test for Eto.Gtk, mscorlib, System. etc.  Not a big deal, though would make it work if used in any platform assembly.

cra0zy:
@cwensley Done.

PR #1036, opened 2018-03-01
[Gtk] Fix Linux NotificationHandler

cra0zy:
Stuff done:
- fix up a crash where the object could get disposed before the activate event occurred 
- don't use generics for Delegate marshaling (does not work with .NET Core it seems)
- create a separate notification object for each notification so that something like this does not happen:
```c#
var n = new Notification();
n.ID = 4;
n.Show();
n.ID = 6;
// activate the existing notification, ID is 6
```


cwensley:
Awesome, thanks for fixing my silly code!

Issue #1041, opened 2018-03-06
Exception thrown when attempting to overwrite ReactiveProperty's value outside constructor


YSRKEN:
Raise an exception when overwrite the value of ReactiveProperty.Value not in constructor method

## Expected Behavior

- When you press "Button 1" button, the black image of width 100 and height 200 displayed in ImageView will be replaced with a black image with width 200 and height 100.
- When you press "Button 2" button, the word "B" is added to the end of the character string displayed on the Label.

## Actual Behavior
  Whichever button you press, the exception "System.Reflection.TargetInvocationException" occurs in mscorlib.dll.

## Steps to Reproduce the Problem

  1. Clone this repository(https://github.com/YSRKEN/EtoApp1)
  2. Build all projects by Visual Studio 2017
  3. Run EtoApp1.Desktop.exe
  4. Press "Button 1" or "Button 2" button

## Code that Demonstrates the Problem

  Please show MainForm.xeto.cs and MainForm.xeto in this repository(https://github.com/YSRKEN/EtoApp1)

## Specifications

  - Version: Eto.Forms v2.4.1, ReactiveProperty v4.2.0, VisualStudio 2017 v15.5.7
  - Platform(s):  Windows 10 Pro v1709 (64bit)

cwensley:
Thanks for reporting the issue! I can't see anything obvious why it isn't working, so I'll have to track that down.

cwensley:
Hey @YSRKEN, sorry for this taking so long, but I finally verified that all other platforms behave the same way where the synchronization context is available after the application is created.

Thanks again!  And thanks @oigami for the fix!

PR #1051, opened 2018-03-15
[WPF] fix for issue #1041: Corresponds to SynchronizationContext at the Application creation stage.

oigami:
In WPF, DispatcherSynchronizationContext is normally set in Application.Run() ( or Dispatcher.Run()). However, if you do a process that depends on SynchronizationContext before that, we will get unintended behavior. So we fixed it to set it at initialization of Eto.Forms.Application.

Although 'Launching a WPF Window in a Separate Thread'[1] talks about a new thread, it is in the same state as immediately after the main function is started.

## Reference
[1] [Launching a WPF Window in a Separate Thread, Part 1](http://reedcopsey.com/2011/11/28/launching-a-wpf-window-in-a-separate-thread-part-1/)

### DIFF
--- a/src/Eto.Wpf/Forms/ApplicationHandler.cs
+++ b/src/Eto.Wpf/Forms/ApplicationHandler.cs
@@ -83,6 +83,9 @@ namespace Eto.Wpf.Forms
 
 		protected override void Initialize()
 		{
+			if (SynchronizationContext.Current == null)
+				SynchronizationContext.SetSynchronizationContext(new DispatcherSynchronizationContext());
+
 			Control = sw.Application.Current;
 			if (Control == null)
 			{


cwensley:
Hey @oigami, thanks for the fix!

I'm not sure if the other platforms set up the context in this way though, have you tried them?

oigami:
I have not tried other platform, but WinForms seems to have WindowsFormsSynchronizationContext in the same way.  According to [WindowsFormsSynchronizationContext.AutoInstall](https://msdn.microsoft.com/ja-jp/library/system.windows.forms.windowsformssynchronizationcontext.autoinstall(v=vs.110).aspx), there may be similar problems because it is set up when creating a Control.

cwensley:
I just tested this on all platforms and they behave the same so this fix should make WPF fall in line.

Thanks again for the fix!

PR #1058, opened 2018-03-18
Remove use of deprecated AppKit API

wjk:
Use of such APIs may be cause for rejection if submitting to the Mac App Store.

Note that I copy/pasted the ampersand-removal algorithm into the three places where it was needed. If you have a better idea of a better, more common location, please let me know. Thanks!


cwensley:
@wjk thank you for submitting this PR! 

If you could put the shared logic into a method in [MacConversions.cs](https://github.com/picoe/Eto/blob/develop/src/Eto.Mac/MacConversions.cs), that would be great!

cwensley:
Looks great, @wjk.  Thanks for the contribution!

PR #1085, opened 2018-04-15
GTK 2/3 Embed PageSetup

halid-durakovic:
I have added few lines of code that allow for selecting page size and orientation for GTK 2 or GTK 3.


halid-durakovic:
I forgot to upload .tt file, and there is no need to separate for GTKCORE api, because current API is same, and these functions exists from GTK2+.

cwensley:
Hey @halid-durakovic, thanks for a the update!  Looks good.  Have you verified that this works on mac and windows?  I'd guess from the API that it would only work on *nix, but hard to know other than trying it out.

cwensley:
Checked it on macOS and works great.  Thanks for the contribution!

Issue #1080, opened 2018-04-12
Support dotnet build


cwensley:
Currently the Mac nugets include targets that aren't compatible with dotnet build (msbuild on .net core) as it doesn't support CodeTaskFactory.  We should pre-compile the tasks instead which should make this work.

SteveGilham:
This is related to [MSBuild Issue #616](https://github.com/Microsoft/msbuild/issues/616); which slaims that this can be resolved by using the [RoslynCodeTaskFactory](https://www.nuget.org/packages/RoslynCodeTaskFactory)
```
<UsingTask  
    TaskName="DoNothing"  
    TaskFactory="CodeTaskFactory"  
    AssemblyFile="$(RoslynCodeTaskFactory)"
    Condition=" '$(RoslynCodeTaskFactory)' != '' ">
```

So I created a blank sample C# Eto Forms app, and added
```
    <PackageReference Include="RoslynCodeTaskFactory" Version="2.0.7" />
```
to the Desktop project, then edited `%USERPROFILE%\.nuget\packages\eto.platform.mac64\2.4.1\build\MacTemplate.targets`  to have
```
	<UsingTask TaskName="_UpdatePList" TaskFactory="CodeTaskFactory"  AssemblyFile="$(RoslynCodeTaskFactory)"
               Condition=" '$(RoslynCodeTaskFactory)' != '' ">
```
When I ran `dotnet build` in the solution directory, that yielded
```
warning : Parameters are discovered through reflection for Type="Class". Values specified in <ParameterGroup/> will be ignored. 
error MSB4065: The "IconFile" parameter is not marked for output by the "_UpdatePList" task.
```

So I removed the `<ParameterGroup> and added a `[Output]` to the `IconFile` property, and the build ran clean.  Doing `dotnet run` in the Desktop project folder then launched the app as expected.

SteveGilham:
As an interim solution, by overriding the `_BuildAppBundle` target, this approach can be fitted into existing build processes without hacking downloaded packages.  This is working for me --

1. Add `<PackageReference Include="RoslynCodeTaskFactory" Version="2.0.7" />` to each project
2. Get amended target file `MacTemplateOverride.targets` from [this gist](https://gist.github.com/SteveGilham/1858f987d410c0c513b93ca2cc6fbace)
[...]

PR #1104, opened 2018-05-13
Addressing Issues 1076 and 1080

SteveGilham:
This is the one about the lack of CodeTaskFactory for `dotnet build`, following the related thread for [MSBuild issue #616](https://github.com/Microsoft/msbuild/issues/616); so if the Roslyn version of the assembly is not present, a warning is emitted, otherwise use it to generate the task.  Manual testing (by patching the nuget package) allowed an application build and launch using `dotnet` (see [comment](https://github.com/picoe/Eto/issues/1080#issuecomment-388638019) on issue #1080)

### DIFF
--- a/build/MacTemplate/MacTemplate.targets
+++ b/build/MacTemplate/MacTemplate.targets
@@ -112,6 +112,9 @@
 		<!--
 			Add any missing keys to the PListFile
 		-->
+		<Warning Text="A reference to the NuGet package RoslynCodeTaskFactory is required in this project" 
+				 Condition=" '$(RoslynCodeTaskFactory)' == '' " /> 
+		
 		<_UpdatePList 
 			PListFile="$(OutputPListFile)"
 			TargetFileName="$(TargetFileName)"
@@ -130,14 +133,8 @@
 		<!-- Set executable bit on launcher if on *nix -->
 		<Exec Command="chmod +x &quot;$(LauncherFileWithPath)&quot;" Condition="'$(OS)'=='Unix'" />
 	</Target>
-	<UsingTask TaskName="_UpdatePList" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll">
-		<ParameterGroup>
-			<PListFile ParameterType="System.String" Required="true" />
-			<TargetFileName ParameterType="System.String" Required="true" />
-			<MacBundleName ParameterType="System.String" Required="true" />
-			<LauncherFile ParameterType="System.String" Required="true" />
-			<IconFile ParameterType="System.String" Output="true" />
-		</ParameterGroup>
+	<UsingTask TaskName="_UpdatePList" TaskFactory="CodeTaskFactory"  AssemblyFile="$(RoslynCodeTaskFactory)"
+               Condition=" '$(RoslynCodeTaskFactory)' != '' ">
 		<Task>
 			<Reference Include="System.Xml" />
 			<Using Namespace="System" />
@@ -161,7 +158,7 @@
                 public string MacBundleName { get; set; }
 				[Required]
                 public string LauncherFile { get; set; }
-
+                [Output]
 				public string IconFile { get; set; }
 
                 XmlDocument xml;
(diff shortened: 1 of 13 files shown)


SteveGilham:
Still TODO would be making this transparent to existing consumers, but adding the RoslynCodeTaskfactory package as a dependency of the eto.platform.mac64.nupkg, and referencing the former directly from the latter (which may be tricky).

SteveGilham:
That does leave the version number in 3 places, though if it's sufficient for purpose there's not great rush to upgrade.

cwensley:
Hey @SteveGilham, thanks a ton for the PR!  This is looking really great!

I'm thinking a better way to resolve this would be to include a compiled version of the tasks.. that way we wouldn't need any extra dependencies.  However, I haven't looked into it yet, but I think it may need separate compiled task assemblies for .net core and for full .net.  Do you know anything about this?

cra0zy:
> I'm thinking a better way to resolve this would be to include a compiled version of the tasks.. that way we wouldn't need any extra dependencies. However, I haven't looked into it yet, but I think it may need separate compiled task assemblies for .net core and for full .net. Do you know anything about this?

I know about this, and yes, you need separate versions compiled, see: https://github.com/cra0zy/MonoGame/tree/mnu/Tools/MonoGame.Content.Builder if you wanna see it in action.

SteveGilham:
Indeed -- even if the code is the same, the linkage information depends on the runtime.  Consuming the Roslyn-based package simply hides that selection.

Having worked in the past of projects that built the build tasks they then consumed, I can say that the build/test/repeat cycle can be a bit painful when the task dlls get locked in processes you don't want to have to kill.

build/MacTemplate/MacTemplate.targets
```diff
@@ -12,10 +12,12 @@
 		<OutputAppPath Condition="'$(OutputAppPath)'==''">$(TargetDir)\$(MacBundleName).app</OutputAppPath>
 		<MacSetDefaultRunConfiguration Condition="'$(MacSetDefaultTarget)' == ''">True</MacSetDefaultRunConfiguration>
 		<MacBuildBundle Condition="$(MacBuildBundle) == '' and $(ProjectTypeGuids) == ''">True</MacBuildBundle>
+		<RoslynCodeTaskProps>$(MSBuildThisFileDirectory)..\..\..\roslyncodetaskfactory\2.0.7\Sdk\Sdk.props</RoslynCodeTaskProps>
```
cwensley: Hm, this is the only line I really have a problem with.  Is this (and the Import below) needed if the nuget package is installed for the project?

cwensley:
Thanks @SteveGilham and @cra0zy for the added information.

I agree that going with the `RoslynCodeTaskFactory` is a good solution at this point and will reduce the overall maintenance of going with a compiled task library.  This PR looks good except the one line that I commented on.  For example, it wouldn't work if the project referenced a different version than 2.0.7.

SteveGilham:
There is that problem, [as I noted earlier](https://github.com/picoe/Eto/pull/1104#issuecomment-388651406) with that package version not being in one definitive location, but in two .nuspec dependency lists and that targets file.

The various cases are 
`RoslynCodeTaskFactory` is used by the consuming project -- then `$(RoslynCodeTaskFactory)` will be set, and whichever version that is will be used directly 
`RoslynCodeTaskFactory` is available as a consequence of being an Eto dependency as per this PR -- then while the .nuspec says 2.0.7 we should be good.
`RoslynCodeTaskFactory` is available on the machine but only in some other version -- well, the build will fail with an advisory message, and with the latest change, `$(RoslynCodeTaskFactoryVersion)` can be set in the consuming project to pick up what's available.

cwensley:
Hey @SteveGilham, thanks for the insight.  I just looked at RoslynCodeTaskFactory, and apparently it already should be including the props file if it's added as a package reference.  

This means those lines that manually import `Sdk.props` from the RoslynCodeTaskFactory package should not even be necessary.

You've already added the dependency to Eto's nuspec so that should take care of it, correct?

SteveGilham:
If the package is included in the project, then the conditional manual inclusion isn't necessary.

Without it, existing projects that don't include RoslynCodeTaskFactory would break, though.  The manual inclusion is purely there to provide backwards compatibility (the package just has to exist on the machine as a dependency of the Eto package for existing builds to just keep on working).  If you're happy to make a breaking release, then that can go.

cwensley:
Thanks, I appreciate that you're thinking about backwards compatibility.  However, I don't think I'd be a breaking change - updating to the new Eto.Mac/Eto.Mac64 nuget packages will automatically pull in the RoslynCodeTaskFactory.

It'd only break when MacTemplate.targets is imported manually, which in that case I don't care for backwards compatibility as it isn't a supported way to use it.  Even so, adding the dependency to your project would fix it which the error message would point them to do.  

I'd rather not have the hack and just keep things as straightforward as possible to ease maintenance going forward (e.g. RoslynCodeTaskFactory could change how it is packaged).

cwensley:
Excellent!  Thanks for the contribution!

PR #1108, opened 2018-05-15
Address Issue 1105 -- AboutDialog localization

SteveGilham:
The simplest thing that could possibly work, but still extensible.
There weren't any other obvious cases of hard-coded strings in the other handlers.

### DIFF
--- a/src/Eto.Gtk/Forms/AboutDialogHandler.cs
+++ b/src/Eto.Gtk/Forms/AboutDialogHandler.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Eto.Drawing;
 using Eto.Forms;
 
@@ -89,6 +89,12 @@ namespace Eto.GtkSharp.Forms
 			get { return Control.WebsiteLabel; }
 			set { Control.WebsiteLabel = value; }
 		}
+
+		/// <summary>
+		/// Gets or sets the text localization.
+		/// </summary>
+		/// <value>The transformation from placeholder string to localized.</value>
+		public Func<string, string> Localize { get; set; } = (x => x);
 
 		public DialogResult ShowDialog(Window parent)
 		{
--- a/src/Eto/Forms/AboutDialog.cs
+++ b/src/Eto/Forms/AboutDialog.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Reflection;
 using Eto.Drawing;
 
@@ -166,6 +166,16 @@ namespace Eto.Forms
 		}
 
 		/// <summary>
+		/// Gets or sets the text localization.
+		/// </summary>
+		/// <value>The transformation from placeholder string to localized.</value>
+		Func<string, string> Localize
+		{
+			get { return Handler.Localize; }
+			set { Handler.Localize = value; }
+		}
+
+		/// <summary>
 		/// Handler interface for the <see cref="AboutDialog"/>
 		/// </summary>
 		public new interface IHandler : CommonDialog.IHandler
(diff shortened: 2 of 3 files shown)


cwensley:
Hm, I wonder if we should include a global mechanism for localizing strings..  There's also a ton of macOS menu items that should allow for translation as well.  Perhaps in Application?  The event should probably include the control its coming from if we move it there though.  Do you have any thoughts on providing global localization instead of just for this single control?

SteveGilham:
I would have to spend more time studying the entire project architecture before opining on a system-wide approach; so far, I simply made the assumption that this project would not want to go down the GTK route and include standard translations for everything -- that way lies maintaining a hundred .resx files, maybe more -- and proposed a purely tactical change for the first component that I spotted this issue in.  

I'm also a great believer in not paying for functionality not used, and starting with the simplest thing that could possibly work, building out only when required.  In this case this proposal places all control with to the end-user, who can in turn simply translate just as much as they need and no more, and arrange the look-up of translated strings to their own convenience; with the default `x => x` if the untranslated English dialect was sufficient to their needs.

In the case of the About Dialog, the implementation I would inject would be like `x => resources.GetResourceString("AboutDialog." + x)`, which leaves the default behaviour undisturbed, and does not impose a look-up scheme on anyone else.

cwensley:
Hey @SteveGilham,

> I'm also a great believer in not paying for functionality not used, and starting with the simplest thing that could possibly work, building out only when required. In this case this proposal places all control with to the end-user, who can in turn simply translate just as much as they need and no more, and arrange the look-up of translated strings to their own convenience; with the default x => x if the untranslated English dialect was sufficient to their needs.

I totally agree with you on this one.  However, I also have to be cognizant of adding API's as removing/changing them would cause ABI breaks (until 3.x).

I am also fortunate that I have some insight into what other issues there are with localization, and I think we can find a good way to address them without going overboard.  I like the approach you did for this control, but I'm thinking of generalizing the `Localize` delegate to the `Application` class instead.  This way, it could be used for all controls, not just this one.  The only caveat is that it should probably be changed to provide the source object (in this case, the AboutDialog) to the delegate as otherwise you wouldn't know what control the localization was for.  I'd recommend using an event and an EventArgs subclass so adding extra parameters to it doesn't break the API.  For example,

```c#
public class LocalizeEventArgs : EventArgs
{
  public string Text { get; set; }
  public object Source { get; private set; }
  internal Initialize(object source, string text)
  {
    Text = text;
    Source = source;
  }
}

public class Application
{
  // events with args are more forward compatible than a Func<> delegate.
  public event EventHandler<LocalizeEventArgs> LocalizeString;

  // so we can override it in an Application subclass
  protected virtual void OnLocalizeString(LocalizeEventArgs e) => LocalizeString?.Invoke(this, e);

  // only use one instance to take it easy on the GC
  LocalizeEventArgs localizeArgs;

  public string Localize(object source, string text)
  {
    localizeArgs = localizeArgs ?? new LocalizeEventArgs();
    localizeArgs.Initialize(source, text);
    OnLocalizeString(localizeArgs);
    return localizeArgs.Text;
  }
  // ...
}
```

Then, in `ThemedAboutDialog`, you'd do:

`Application.Instance.Localize(this, "Developers:")`

and to provide your own localization:

`Application.Instance.LocalizeString += (sender, e) => e.Text = resources.GetResourceString(e.Source.GetType().FullName + "." +e.Text);`

What do you think?

SteveGilham:
That would work.  Not an approach I'd considered before.

cwensley:
Thanks again for the PR, I've implemented what I suggested instead of merging this one.  Your contribution is much appreciated!

PR #1125, opened 2018-06-09
Added constructor for UITimer that provides a delegate to handle the …

allsorts46:
…Elapsed event.

Added overload of Color.Blend() which accepts a separate blending factor.


allsorts46:
A couple of small conveniences that I added because I found them useful whilst working with Eto. This is mainly to serve as 'My First Pull Request' so I get to see how the process works. Changes are only to core classes, nothing platform-specific.

cwensley:
Looks really good, thanks for submitting the PR!

PR #1133, opened 2018-06-16
Gtk: Fix missing Application.Restart()

MartinRothschink:
This fixes the missing implementation of Restart() in Gtk.

### DIFF
--- a/src/Eto.Gtk/Forms/ApplicationHandler.cs
+++ b/src/Eto.Gtk/Forms/ApplicationHandler.cs
@@ -61,7 +61,7 @@ namespace Eto.GtkSharp.Forms
 			GLib.ExceptionManager.UnhandledException -= OnUnhandledException;
 			Gtk.Application.Quit();
 
-			// TODO: restart!
+			Process.Start(Assembly.GetEntryAssembly().Location);			
 		}
 
 		string badgeLabel;
--- a/src/Eto.Gtk/Forms/ApplicationHandler.cs
+++ b/src/Eto.Gtk/Forms/ApplicationHandler.cs
@@ -61,16 +61,7 @@ namespace Eto.GtkSharp.Forms
 			GLib.ExceptionManager.UnhandledException -= OnUnhandledException;
 			Gtk.Application.Quit();
 
-			var cmdLine = string.Empty;
-			var oldArgs = Environment.GetCommandLineArgs();
-			if (oldArgs.Length > 1)
-			{
-				var args = new String[oldArgs.Length - 1];
-				Array.Copy(oldArgs, 1, args, 0, args.Length);
-				cmdLine = String.Join(" ", oldArgs);
-			}
-
-			Process.Start(Assembly.GetEntryAssembly().Location, cmdLine);			
+			RestartInternal();			
 		}
 
 		string badgeLabel;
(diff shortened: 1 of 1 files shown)


cra0zy:
You need to take into account that this might be running on Mac/Linux so you need to do extra steps like to figure out what command was called to start the app.

MartinRothschink:
I tested it on Linux with Mono 5.12 but yes, command line arguments are not preserved.

cra0zy:
> I tested it on Linux with Mono 5.12 but yes, command line arguments are not preserved.

No, it is not guaranteed to work the way you "tested" it, let me elaborate:
 - If `mono` is not set as default application, but instead you used something like `wine` or archive manager set as default for .exe files, the app will not restart
 - you need to detect what your app is using mono or dotnet and call the appropriate 
 - you need to detect if your app is bundled and call the appropriate executable in that case

So there is a lot that you'd need to cover to implement this.

MartinRothschink:
Harry, I'm not restarting the entire runtime (in which case you're right), I'm restarting the managed application  while the runtime is still loaded:

Process.Start() is called to execute the entry assembly again. Therefore it should not matter if Mono or .Net Framework is currently used.
It should not matter if there is a default application to execute .exe files. 
Also an app bundle should not matter because the path to the entry assembly is still the same.

cra0zy:
> Process.Start() is called to execute the entry assembly again. Therefore it should not matter if Mono or .Net Framework is currently used.

No, `Process.Start()` is calling stuff like mimeopen, ie. you can open up a .jpg file with it and it will start the default application for it.

I will not bother to correct the rest of your post since the above statement shows the flaws with the rest of your logic.

MartinRothschink:
Yes, Process.start() attempts to run the default application for a file extension. But .exe is handled differently in my experience.

On my system, the default app for .exe is archive manager. Process.Start() does **not** run archive manager but runs the application again.

I have tested the change on some systems and think it is an improvement on the current situation where Restart() simply terminates the application.

Harry, please give it a try.

Regards
Martin

cra0zy:
> But .exe is handled differently in my experience.

Its not: https://github.com/mono/mono/blob/master/mcs/class/System/System.Diagnostics/Process.cs

Tbh, you don't know how to fix it for Unix systems, so just add `if Windows` clause and leave a `TODO` for Unix platforms.

MartinRothschink:
Harry, that's the high level part, the real work is done here:
https://github.com/mono/mono/blob/master/mono/metadata/w32process-unix.c

Have a look at line 1848:

```
	/* Check for CLR binaries; if found, we will try to invoke
	 * them using the same mono binary that started us.
	 */

	if (is_managed_binary (prog)) {
```

Martin

cra0zy:
> Harry, that's the high level part, the real work is done here:
> https://github.com/mono/mono/blob/master/mono/metadata/w32process-unix.c

Interesting... tho its such a bad idea to do that...

From what I can tell .NET Core is not doing it so it seems to be mono only quirk: https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Unix.cs#L298

Tho there are other stuff you need to take into account like mkbundle and dotnet publish as well.

MartinRothschink:
Updated to handle all your requirements. Tested with Mono and mkbundle on Linux, dotnet run and publish on Windows and Linux and traditional .Net Framework.

Issue #1141, opened 2018-06-27
HighDPI: different scaling of Graphics in WinForms and Direct2D engines.


e673:
I'm developing a DPI-aware application which has a `Drawable` element with custom `OnPaint` implementation.

In `OnPaint` method I apply scaling transform to `e.Graphics` in order to get one-by-one correspondence between points and physical pixels. I get the ratio of scaling transform from `Screen.LogicalPixelSize` property.

Everything works fine with Wpf, WinForms, Gtk engines but I've got a problem with Direct2D engine.
When using Direct2D engine, I receive scaled `e.Graphics` object and have to calculate the exact scale ratio as a ratio between control size and clipping rectangle size.

Consider a control with a width of 2390. In `OnPaint` method I receive `e.Graphics` instance and apply scaling transform corresponding to actual scaling factor. I expect a clipping rectangle `(0, 0, 2390, *)` but got `(0,0, 2388, *)`.

I've analyzed `Eto.Direct2D.GraphicsHandler.ClipBounds` property. The problem seems to be in rounding and errorneous clip rectange calculation in `Eto.Direct2D.GraphicsHandler`.

1. The values are floating point but they are rounded in the middle of calculations in `ToEto()` method.
Changing `ToEto` implementation to the following
```cs
public static PointF ToEto(this s.Vector2 value)
{
    return new PointF(value.X, value.Y);
}
```
solves the rounding problem.

2. Calculation of new rectangle size using `EndLocation` is also wrong:
`(Width - 1) * scale + 1 != Width * scale`.

The correct implementation of `ClipBounds` property should be:

```cs
public RectangleF ClipBounds
{
	get
	{
		if (Control != null)
		{
			// not very efficient, but works
			s.Matrix3x2 transform = currentTransform; // Control.Transform; - the result should not depend on PixelOffsetMode
			transform.Invert();
			var start = s.Matrix3x2.TransformPoint(transform, clipBounds.Location.ToDx()).ToEto();
			var end = s.Matrix3x2.TransformPoint(transform, clipBounds.BottomRight.ToDx()).ToEto();
			return new RectangleF(start, new SizeF(end.X - start.X, end.Y - start.Y));
		}
		else
			return new RectangleF();
	}
}
```

e673:
[...]

PR #1143, opened 2018-06-27
DPI scaling fix in Direct2D engine, ClipBounds property fix in Direct2D and WinForms

e673:
It's a fix to an issue #1141.

1. Direct2D: `ClipBounds` value was incorrect when DPI scaling was on due to rounding and logical problems.
2. WinForms: `ClipBounds` should not depend on PixelOffsetMode value.
3. Direct2D: since the control size is physical size regardess of screen DPI, there is no need to scale WicRenderTarget.

### DIFF
--- a/src/Eto.WinForms/Drawing/GraphicsHandler.cs
+++ b/src/Eto.WinForms/Drawing/GraphicsHandler.cs
@@ -427,8 +427,11 @@ namespace Eto.WinForms.Drawing
 		{
 			get
 			{
+				var oldValue = this.Control.PixelOffsetMode;
 				this.Control.PixelOffsetMode = sdd.PixelOffsetMode.None;
-				return this.Control.ClipBounds.ToEto();
+				var result = this.Control.ClipBounds.ToEto();
+				this.Control.PixelOffsetMode = oldValue;
+				return result;
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


src/Eto.WinForms/Drawing/GraphicsHandler.cs
```diff
@@ -425,7 +425,11 @@ public IMatrix CurrentTransform
 
 		public RectangleF ClipBounds
 		{
-			get { return this.Control.ClipBounds.ToEto(); }
+			get
+			{
+				this.Control.PixelOffsetMode = sdd.PixelOffsetMode.None;
+				return this.Control.ClipBounds.ToEto();
```
cwensley: I think you need to restore the previous mode after getting the bounds here.

cwensley:
Hey @e673, this is great thanks for the PR!  Looking good except for restoring the PixelOffsetMode on WinForms when getting the ClipBounds.

e673:
> Hey @e673, this is great thanks for the PR! Looking good except for restoring the PixelOffsetMode on WinForms when getting the ClipBounds.

Yes, I was thinking about it. But this value is updated in every call of any drawing method (DrawLine, FillRectangle etc) by calling `SetOffset(...)`. So I see no sense in restoring its value.

cwensley:
Awesome, thanks for the contribution!!

Issue #1187, opened 2018-08-27
App Closes unexpectedly while Main is hidden.


kedrin:
## Expected Behavior
Application continues running while main form is hidden

## Actual Behavior
Application exits when main form is hidden and secondary window is closed. This occurs on GTK3 on linux and Windows, Wpf operates as expected. This occurs even when the application main form is set or started with app.run(new Form());

## Steps to Reproduce the Problem
On the GTK3 platform with a tray indicator icon the main window is hidden on close, a dialog is opened through the tray icons context menu. When closing the dialog the application exits as if the main window was closed.

## Code that Demonstrates the Problem

https://github.com/kedrin/ETOAppClosureError

  - Version: 4.2.1
  - Platform(s): WPF, Gtk3
  - Operating System(s):  Windows 10 Ubuntu 18.04

cwensley:
Thanks for reporting the issue!

MartinRothschink:
This is a duplicate of my issue #1132

Issue #1132, opened 2018-06-14
[GTK] Dialog close closes invisible main window


MartinRothschink:
Still working on my tray application with an invisible main  window. I added a separate about dialog which is created from the tray context menu. Closing this dialog closes the entire application on GTK. MAC and WPF are fine.

## Expected Behavior
Closing my about dialog should not close the application if main window is invisible.

## Actual Behavior
Closing my about dialog dialog closes invisible main window and application.

## Steps to reproduce the issue

1. Run code below, only tray icon is visible.
2. Click on tray icon to open context menu, select About...
3. About dialog is displayed, close dialog 
4. Application exists

## Code that Demonstrates the Problem

```
using System;
using Eto.Forms;
using Eto.Drawing;

namespace EtoTrayApp
{
   public class About : Dialog
   {
      public About()
      {
         Title = "My About Dialog";
         ClientSize = new Size(400, 350);

         Content = new StackLayout
         {
            Padding = 10,
            Items =
            {
               "About!",
            }
         };
      }
   }

   public class MainForm : Form
   {
      private bool _startup = true;
      private TrayIndicator _tray;
      public MainForm()
      {
         Title = "My Eto Form";
         ClientSize = new Size(200, 200);
         _tray = new TrayIndicator
         {
            Image = Eto.Drawing.Icon.FromResource("EtoTrayApp.error.ico"),
            Menu = new ContextMenu
            {
               Items =
               {
                  new ButtonMenuItem {Text = "About...", Command = new Command((s, e) =>
                  {
                     var about = new About();
                     about.ShowModal(this);
                  })},
                  new ButtonMenuItem {Text = "Quit", Command = new Command((s, e) => Application.Instance.Quit())}
               }
            }
         };

         ShowInTaskbar = false;
         _tray.Visible = true;
      }

      protected override void OnShown(EventArgs e)
      {
         base.OnShown(e);
         if (_startup)
         {
            // change this to false to show the issue
            Visible = false;
            _startup = false;
         }
      }
   }
}
```

## Specifications

  - Version: Eto.Gtk 2.4.1
  - Platform(s): Linux Gtk
  - Operating System(s): Ubuntu 18.04 with Mono 5.12

MartinRothschink:
Sample project https://1drv.ms/u/s!Ajb182QihKkLs8h16lv-1MzWq0H30A

cra0zy:
You should not be allowing the window to get closed, this is normal...

[...]

PR #1233, opened 2018-10-04
[GTK] Fix crash in Dialog.ShowModal(). Fix closing of hidden main window.

MartinRothschink:
Addresses issue #1132 and #1187


MartinRothschink:
Mac upload fails, nothing I can fix from my side.

cwensley:
Hm, really not sure why the CI build is failing, but this PR looks good to me!

Thanks for the contribution!

cwensley:
Just a note that I found that with this update GTK applications never closed if you opened (then closed) a secondary dialog or window then closed the main window.  I've fixed the issue and hopefully it hasn't affected this issue as well.  

Setting `Window.Visible = false` should flag the window differently than if it was closed.  I've done some testing you you may also want to do further testing to make sure it works in your scenario as well.

Cheers!
Curtis.

PR #1240, opened 2018-10-25
Fix using font variants with RichTextArea

cwensley:
This includes fonts with Light/UltraLight/Medium/Strong/Narrow variants.  WPF does not by default save/load these from RTF correctly, so we transform the font names when saving or loading.

### DIFF
--- a/src/Eto.Mac/Drawing/EtoFontManager.cs
+++ b/src/Eto.Mac/Drawing/EtoFontManager.cs
@@ -1,0 +1,78 @@
+using System;
+#if XAMMAC2
+using AppKit;
+using Foundation;
+using ObjCRuntime;
+using CoreAnimation;
+using CoreImage;
+#elif OSX
+using MonoMac.AppKit;
+using MonoMac.Foundation;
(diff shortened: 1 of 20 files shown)


msasso69:
Curtis, I'm puzzled by the change you have done in 

"Fix NRE when setting DropDown.DataStore to null",

I refer to the line in src/Eto/CollectionChangedHandler.cs:362:
```
return Collection?.IndexOf(item) ?? -1;
```
in particular to the second part:
```
IndexOf(item) ?? -1;
```
IndexOf(item) returns either -1 or a position greater or equal to zero, why checking whether it is null?
...and the operator ?? does not protect against item == null either

Did you mean something else or am I missing something?
Hope I'm not wasting your time

cwensley:
@msasso69, I also added the question mark after Collection. So ‘Collection?.IndexOf(item)’ returns a nullable int (int?). If Collection is null, it will return -1 instead of crashing.

msasso69:
oh! right! I didn't notice at first! shame on me

PR #1255, opened 2018-11-15
Implemented some properties of Eto.WinForms.Forms.HwndFormHandler

kike-garbo:


cwensley:
This is great, thanks for the contribution!  This might even fix a few issues on my list. (;

PR #1281, opened 2018-12-07
Implemented a style for WPF/WInForm to set the notification icon.

MartinRothschink:
On WPF/WinForms, the ContentImage property does nothing.  Added a Style to set one of the predefined icons.


cwensley:
Hey @MartinRothschink this is great, thanks again!

I don't necessarily like having the `NotificationIcon` enum redefined, though doing it this way will allow us to change the implementation if we need to in the future.

PR #1292, opened 2018-12-25
Add Eto.WinForms and Eto.Wpf for .NET Core 3.0

fredrikhr:
## Summary

* Includes the changes described in #1410 
* Adds .NET Core 3.0 TFM to `Eto.WinForms` and `Eto.Wpf` projects.
* Adds `Eto.WinForms` and `Eto.Wpf` to `Eto.Core.sln` solution file
* For maintanence of .NET Core WinForms support it might be advisable to follow the [Issues for the dotnet/winforms repository](https://github.com/dotnet/winforms/issues).

## Issues

* Using the `Microsoft.NET.Sdk.WindowsDesktop` SDK in the project file, causes loading errors in Visual Studio 2017 (and lower). The new .NET Core 3.0 can currently only be loaded in Visual Studio 2019 (and above).
   @cwensley voiced that this is okay.
* `Eto.WinForms` and `Eto.Wpf` both depend on the `Windows7APICodePack-Shell` NuGet package which does not target any version of .NET Standard or .NET Core. `Eto.Wpf` additionally depends on the package `Extended.Wpf.Toolkit`. Because of this a `NU1701` warning is generated and explicitly supressed in the project files. Thus far I have not experienced any problems with these dependencies in .NET Core, even though they target full framework.
   We will have to keep an eye out for a `Windows7APICodePack-Shell` and `Extended.Wpf.Toolkit` package targeting .NET Standard in the future.


cwensley:
Hey @couven92, first of all big thanks for putting this together!  Here are my thoughts:

- We can't use multi-targeting unfortunately until VS for Mac supports this properly.  It mostly works, but it will break any .net core app in the sln such as Eto.Test.Gtk as VS for Mac only recognizes (and builds) the first framework listed.  There has been some work on this lately that I've noticed, so hopefully soon we can use it.
- We shouldn't need a separate NuGet package or platform as it will depend on what framework you are running on.  I think we can go ahead and add the .net core version of Eto.WinForms.dll to the existing Eto.Platform.Windows nuget package, as it should not affect existing projects.
- I don't mind adding the new project(s) to the .sln.  Having a few messages show up when loaded in VS 2017 isn't a big deal.  If necessary we can add a separate VS 2017 sln.  Also, have you tested this on VS for Mac?
- I'm hoping a version of Windows7APICodePack-Shell will pop up that can be used from .net core, in the meantime your solution to avoid the warning is perfect.
- We will need to get CI working, and I see you've attempted a few things.  It appears it's hanging up on the nuget restore, possibly due to an old version of nuget.exe?

fredrikhr:
@cwensley 
> We shouldn't need a separate NuGet package or platform as it will depend on what framework you are running on. I think we can go ahead and add the .net core version of Eto.WinForms.dll to the existing Eto.Platform.Windows nuget package, as it should not affect existing projects.
* I merged the .NET Core projects into the original Eto projects.
> We can't use multi-targeting unfortunately until VS for Mac supports this properly. It mostly works, but it will break any .net core app in the sln such as Eto.Test.Gtk as VS for Mac only recognizes (and builds) the first framework listed. There has been some work on this lately that I've noticed, so hopefully soon we can use it.
* Okay, I used multi-targeting, but always put the .NET Framework targets first.
> I don't mind adding the new project(s) to the .sln. Having a few messages show up when loaded in VS 2017 isn't a big deal. If necessary we can add a separate VS 2017 sln. Also, have you tested this on VS for Mac?
> I'm hoping a version of Windows7APICodePack-Shell will pop up that can be used from .net core, in the meantime your solution to avoid the warning is perfect.
* The Pull Request introduction comment has been updated.
> We will need to get CI working, and I see you've attempted a few things. It appears it's hanging up on the nuget restore, possibly due to an old version of nuget.exe?
* I got the Azure CI to work, I'll look into AppVeyor. Updating `nuget.exe` to newest might resolve the issue.

fredrikhr:
Sorry for taking so long with this, I only yesterday had time to look into this again.

So I cleaned up a bit, and I am still trying to figure out how to get the CI to pass. Suggestions are very welcome. I have been trying out a lot of different things, but so far I have been unsuccessful. To aid my diagnostic of what actually happens during the CI build, I get MSBuild to produce binary-logs and publish them (I'll remove that later if unwanted).

So here's what I found:
* CI on Mac correctly picks up on the installed .NET Core 3.0 SDK, but then fails because of mismatch between newer and old versions of MSBuild support libraries. ref.: https://rider-support.jetbrains.com/hc/en-us/articles/360004180039-Incompatible-versions-of-Mono-MSBuild-and-NET-Core-SDK-MSB4018- and https://github.com/NuGet/Home/issues/7956
* CI on Windows (both Azure Pipelines and AppVeyor) do not pick up on the new installed .NET Core SDK which is very strange, because Visual Studio shipped MSBuild will happily do so locally on my computer.

One possible way would to run `dotnet msbuild` instead of Visual Studio shipped `MSBuild`. However, we use `CodeTaskFactory` in several of the build project files and we'd have to convert all of them to .NET Core supported targets.

/cc @cwensley

fredrikhr:
In an attempt to address the CI issues, I have changed the CI-builds to use the .NET Core CLI `dotnet msbuild` command instead of the MSBuild command shipped with Visual Studio. Because of this I have changed the usages of `CodeTaskFactory` tasks to use the .NET Core supported `RoslynCodeTaskFactory`.

fredrikhr:
Great, now CI is, in theory, able to build Eto. The reason it fails now is because of References. I'll continue to .NET Core-ify the remaining projects until it works...

@cwensley maybe you'd like me to do the full move to .NET Core build in a separate PR? Because once CI using the `dotnet` CLI works, porting WinForms (and probably WPF) becomes trivial. And it would be a good future move

cwensley:
@couven92 looking at the build logs, it also appears to be failing on Xamarin.Mac on mac.  This is because Xamarin.Mac does not support using dotnet cli (and it will probably be a while for it to be supported).

I think holding off using dotnet CLI for CI would be best here (too many roadblocks).  However, we can add any of the projects that support .NET Core to Eto.Core.sln so you can use dotnet CLI on that solution manually.

fredrikhr:
🎉 Yes! Now everything is finally working! I ended up installing the .NET Core 3.0 Preview 5 SDK and then I did a little trick where I teach MSBuild on the build agent about the new SDK.

That way, the entire build is still performed by the MSBuild that is installed on the agent and does all the full framework stuff. But since MSBuild now knows about the new SDK, it is also able to compile Eto.WinForms for the `netcoreapp3.0` TFM using the new `Microsoft.Net.Sdk.WindowsDesktop`. Because of loading issues with existing MSBuild binaries and newer version of these packages, I disabled the WindowsDesktop SDK on the Mac CI build. (It doesn't really make much sense compiling for WindowsDesktop on Mac anyways).

@cwensley I think we are ready for a review! :)

cwensley:
Hey @couven92, that is excellent news!!  Thanks so much for doing all the groundwork on this, I'm excited to get it in!

I'm currently on a deadline for the end of the month, so it may take a short while for me to look at this and merge it in, but on first pass it looks great!

fredrikhr:
@cwensley I now also managed to get WPF working for .NET Core 3.0. the changes for WPF are very similar to those I made for WinForms.

cwensley:
Hey @couven92, thanks for keeping this PR up to date, it is really appreciated.  I had some time to look into it and loading it in VS for Mac gives me these errors when trying to restore packages (which aborts the nuget restore process for the rest of the solution):

```
/usr/local/share/dotnet/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/curtis/Projects/McNeel/rhino/src4/DotNetSDK/Eto/src/Eto.WinForms/Eto.WinForms.csproj]

/usr/local/share/dotnet/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/curtis/Projects/McNeel/rhino/src4/DotNetSDK/Eto/src/Eto.Wpf/Eto.Wpf.csproj]

/usr/local/share/dotnet/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/curtis/Projects/McNeel/rhino/src4/DotNetSDK/Eto/test/Eto.Test.Wpf/Eto.Test.Wpf.csproj]

/usr/local/share/dotnet/sdk/3.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/curtis/Projects/McNeel/rhino/src4/DotNetSDK/Eto/test/Eto.Test.WinForms/Eto.Test.WinForms.csproj]
```

Perhaps we have to set `HaveWindowsDesktopSdk` to false on the condition that you are building on !windows perhaps?

fredrikhr:
@cwensley I added the check for Windows. Turns out that MSBuild has built-in functions for that. 😀 

The CI Build fails now, but it looks like the error comes from the agent not being able to install GTK#, can you look into that?

cwensley:
Thanks again!  I've incorporated your changes into the develop branch, thanks for the work you put into this.  Please check out the changes I've done and let me know if there's anything else missing.

PR #1295, opened 2019-01-06
Make Eto buildable with .NET Core SDK

fredrikhr:
In the past, Eto defines the `_GetCommonFiles` and `GenerateVersion` tasks in `build\Common.Build.targets` by using the MSBuild `CodeTaskFactory` inline-tasks.

`CodeTaskFactory` inline-tasks are not supported by the .NET Core SDK, although Microsoft is working on the `RoslynCodeTaskFactory` to replace the older `CodeTaskFactory`, but as I tried to use that, I ran into issue https://github.com/Microsoft/msbuild/issues/3726 which makes this alternative unusable as of now. Anyways, using `RoslynCodeTaskFactory` would not work on .NET Core SDKs using MSBuild with versions lower than 15.8.

As an alternative to using `CodeTaskFactory`, I simply wrote the task in a regular C# project referencing the `Microsoft.Build.Framework` package. `Common.Build` now loads the `_GetCommonFiles` and `GenerateVersion` from the compiled assembly from the new `Eto.Build.Tasks` project instead.

`Eto.Build.Tasks` drops its binaries in `artifacts\build\$(Configuration)\$(TargetFramework)` and a custom target copies the generated binary to `artifacts\build` from where it is imported in `Common.Build`.

MSBuild tasks target either `net46` or `netstandard2.0`. In order for all Eto projects (including those who target .NET Standard version less than 2.0) to target `Eto.Build.Tasks` (as declared in `Common.Build.props`), `netstandard1.0` has been added to `Eto.Build.Tasks` as a third `TargetFramework`. However, making Eto projects reference `Eto.Build.Tasks` is really only for MSBuild, it does not matter to anyone that the `netstandard1.0` version of `Eto.Build.Tasks` is completely empty.


turbo:
KUTGW

fredrikhr:
Closing PR in favour of #1296

PR #1296, opened 2019-01-06
Make Eto buildable with .NET Core SDK

fredrikhr:
In the past, Eto defines the `_GetCommonFiles` and `GenerateVersion` tasks in `build\Common.Build.targets` by using the MSBuild `CodeTaskFactory` inline-tasks.

However, while I was working on #1295 I noticed the following:

* The `_GetCommonFiles` task is not used anywhere the .NET Core SDK is involved, so it can be ignored.
* The `GenerateVersion` task calculates the the Revision and Build numbers from an existing version. You can do it using static properties and MSBuild Property Functions. The expression is a bit hacky, but it works really well, and thus removes the need to rely on `CodeTaskFactory`.

By using Property Functions instead of Custom Code-Task-Factory-inline-Tasks, Eto can with these changes be built by the .NET Core SDK without any further changes.

Although the MSBuild expression is hacky, this incurs far less overhead than the solution proposed in #1295 and I suggest using this solution instead of #1295.


cwensley:
Awesome!  

I hope that we'll be able to use `CodeTaskFactory` or equivalent in the future with dotnet build tools, but until then this is a great workaround.

Thanks for making the changes!  Hopefully this will make developing for Eto less painful for some.

PR #1311, opened 2019-02-01
Can't find fonts with face suffixes in their 'Source' property 

acormier:
(ex: DINPro-Regular) so remove -Regular suffix in those cases


cwensley:
Awesome!  Fix looks great.

PR #1336, opened 2019-03-04
New SegmentedButton and ToggleButton

### LINKED ISSUE #889: Workaround to group Eto-Buttons in MacOS ?
I would like to group some buttons in MacOS like described here:
[https://christiantietze.de/posts/2016/06/segmented-nstoolbaritem/](https://christiantietze.de/posts/2016/06/segmented-nstoolbaritem/)

My idea is to configure the grouping with the text-property, which I don't need, and add a style:
```
var button1= new Button { Style="groupedtoolbarbutton", Text="g1", Image= img1 };
var button2= new Button { Style="groupedtoolbarbutton", Text="g1", Image= img2 );
var button3= new Button { Style="groupedtoolbarbutton", Text="g2", Image= img3 };
var button4= new Button { Style="groupedtoolbarbutton", Text="g2", Image= img4 );
```
In the style-handler I would collect the buttons and group them according to the text-property.
```
Eto.Style .Add <Eto.Mac.Forms.Controls.ButtonHandler >("groupedtoolbarbutton", handler =>
{
        var group = handler.Text;
        handler.Text = string.Empty;
	// collect buttons and join buttons with equal group
	// todo
});
```
I just don't know, how... I'm absolute not familiar with MacOS-UI and Cocoa, that's why I'm using Eto. Does somebody maybe know, if  this is possible and how to put this in c# in the style-handler?

Thanks
Tom

cwensley:
- Added ToggleButton, which is a Button, but can be toggled on an off.
- Added SegmentedButton which allows you to have a button with different segments, and items with a menu.
- Added BindableExtensions.Inverse() for bool and bool? bindings to more easily invert its value.
- Added BindableExtensions.DefaultIfNull() for value and class types so you can easily convert a nullable binding to non-nullable.
- Button.Size should at least be the MinimumSize, but not if it is -1.
- Add Control.OnApplyCascadingStyles and ApplyStyles for controls that want to style its children when they are not a Container.
- Fix issue applying cascading styles with the Handler class.
- Added PropertyStore.TrySet to more efficiently execute logic when the value is changed.
- Fixed PropertyStore.Set with the property changed delegate parameter when the default value is different.
- Wpf: Setting relative position of menu now places from upper left corner unless showing exactly below or to right of control.
- Wpf: Release capture of the mouse when handling Control.MouseUp but not MouseDown.
- Gtk: Fix converting relative point to screen for Button

Fixes #889

### DIFF
--- a/src/Eto.Gtk/Eto.Gtk.csproj
+++ b/src/Eto.Gtk/Eto.Gtk.csproj
@@ -4,6 +4,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+    <RootNamespace>Eto.GtkSharp</RootNamespace>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DefineConstants>TRACE;PCL;GTK3;DEBUG;NETSTANDARD2_0;GTKCORE</DefineConstants>
--- a/src/Eto.Gtk/Eto.Gtk2.csproj
+++ b/src/Eto.Gtk/Eto.Gtk2.csproj
@@ -287,6 +287,7 @@
     <Compile Include="Forms\TableLayoutHandler.gtk2.cs" />
     <Compile Include="CustomControls\BaseComboBox.gtk2.cs" />
     <Compile Include="CustomControls\BaseComboBox.gtk3.cs" />
+    <Compile Include="Forms\Controls\ToggleButtonHandler.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
   <ProjectExtensions>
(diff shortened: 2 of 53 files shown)


harry-cpp:
Interesting, interesting indeed.

Can you give me an example of segmented button usage?

cwensley:
@cra0zy, check out the [sample](https://github.com/picoe/Eto/blob/develop/test/Eto.Test/Sections/Controls/SegmentedButtonSection.cs) and [unit tests](https://github.com/picoe/Eto/blob/develop/test/Eto.Test/UnitTests/Forms/Controls/SegmentedButtonTests.cs)

A simple example is:

```c#
var sb = new SegmentedButton { Items = { new ButtonSegmentedItem { Text = "Item 1" } /*...*/ } };
sb.ItemClick += (sender, e) => Console.WriteLine($"Item: {e.Item}");
```

macOS uses NSSegmentedControl, whereas all other platforms simply use a TableLayout with the ToggleButton.  However, with Gtk 3.x it styles it with a 'linked' class so they appear together which appears to be the [recommended way](https://wiki.gnome.org/HowDoI/Buttons) to do that.

harry-cpp:
> However, with Gtk 3.x it styles it with a 'linked' class so they appear together which appears to be the [recommended way](https://wiki.gnome.org/HowDoI/Buttons) to do that.

Yup, I can confirm that. I need to think of an API to implement headerbars now :thinking:

PR #1363, opened 2019-04-10
Add other configurations for vscode

cwensley:
- Can now run Wpf on windows and no longer prefers 32 bit so it can be debugged
- Mac64 and Gtk2 on Mac or Linux both run using a modified vscode-mono-debug extension


cwensley:
To use vscode for Mac64, you need to use the 'extended' mono-debug extension from [here](https://github.com/cwensley/vscode-mono-debug/releases)

PR #1410, opened 2019-05-22
Build optimizations, future proofing and refactorings

fredrikhr:
This is a little update to help improve and modernize the CI builds for the repository.

* Update nuget.exe binary
   Updates `build/nuget.exe` to lastest stable version at time of writing (version `5.0.2`)
* The `licenseUrl` element in NuSpec is being deprecated and triggers a NU5125 warning.
  The solution is to use the `license` element instead and link to the `LICENSE.txt` that we embed in the NuGet.
* Adds `<AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>` to some project files to address MSBuild warnings on version number mismatch of referenced assemblies.
* Adds arguments, commands and build steps to CI builds to cause MSBuild to emit binary logs.
  If the build fails, binary log files will be published as build artifacts for later review using tools like [MSBuild Structured Log Viewer](https://github.com/KirillOsenkov/MSBuildStructuredLog)
  This means that we can reduce the MSBuild verbosity to minimal. This results in shorter build logs, and because of that shortened build time.
* Add `MSBuildTaskHelper.props` which implements smart logic to select proper `TaskFactory` and `AssemblyFile` attribute values in inline-code MSBuild tasks.
  This lays the ground for future usages of the build pipeline using .NET Core tooling (as .NET Core requires `RoslynCodeTaskFactory` to be used)
* Replace the nuspec files in `build/nuspec` with project property files for SDK-style projects.
  Using the SDK to produce the nuspec and package the NuGet package is less error-prone as the SDK correctly infers all dependencies and nuspec contents. In the future this will make creating multi-targeting NuGet packages a lot easier, since the SDK automatically picks up the targeted TFMs from the project file.
* Instead of using `nuget.exe`, use the MSBuild `Restore` target.
  Letting MSBuild figure out how to restore is better than reling on an external tool. This also reduces the need to keep the `nuget.exe` binary in `build/nuget.exe`. Finally, since MSBuild is called within its own process, errors and warnings are properly propagated to the error and warning system of the CI build agent.
  Using the `Restore` target has one gotcha when targeting `.netstandard2.0` (as we do with `Eto.Gtk`). As discussed in https://github.com/Microsoft/msbuild/issues/2455#issuecomment-323575719 you have to force MSBuild to re-evaluate the project file after restore is complete. This can easily be done by adding a unique global property (`_RestoreEnforceEvaluate=True`) for the restore target.
  Ideally, this will in the future remove the need to use `packages.config` files and will enable us to move everything over to `PackageReference` tags in the project files.


fredrikhr:
These changes were part of the modifications I did while I was porting Eto.WinForms to .NET Core 3.0. However, the changes are general in nature and can be discussed seperately.

By the way, I also noticed that even non-SDK-style projects can use `PackageReference` tags and the `Restore` target. Probably this means that they also support the `Pack` target. This means that we could move entirely away from the nuspec files and pack all NuGets using the same mechanism.

@cwensley what do you think?

PR #1412, opened 2019-05-23
Handle `Enter` key only when `EtoDataGrid` is in edit mode

### LINKED ISSUE #1411: Activated event of the TreeGridView is not triggered on WPF
On WPF, `Activated` event of the `TreeGridView` is not triggered when the user presses the `Enter` key.

## Expected Behavior
`Enter` key should trigger `Activated` on `TreeGridView`

## Actual Behavior
Nothing happens. A closer look shows that `EtoDataGrid` class marks the event as handled before `Activated` is triggered. See:
https://github.com/picoe/Eto/blob/34c70098575bd3560b584f907d1bb717034e2f14/src/Eto.Wpf/Forms/Controls/GridHandler.cs#L43

## Steps to Reproduce the Problem

  1. Run Eto test application
  2. Go to TreeGridView section
  3. Focus on a row and hit `Enter`

  - Version: 2.4.1
  - Platform(s): WPF
  - Operating System(s): Windows 10

brtn:
Before this commit, `EtoDataGrid` handled the `Enter` key unconditionally.
This caused side effects such as `Activated` event on `TreeGridView` not
being triggered. With this change, we check whether we are in edit mode or
not and handle only if we are actually dealing with an edit.

Should fix #1411

### DIFF
--- a/src/Eto.Wpf/Forms/Controls/GridHandler.cs
+++ b/src/Eto.Wpf/Forms/Controls/GridHandler.cs
@@ -8,6 +8,7 @@ using swm = System.Windows.Media;
 using swi = System.Windows.Input;
 using Eto.Forms;
 using System.Collections;
+using System.ComponentModel;
 using Eto.Wpf.Forms.Menu;
 using Eto.Drawing;
 using Eto.Wpf.Drawing;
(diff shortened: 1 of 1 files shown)


cwensley:
Hey @brtn, thanks for submitting the issue and the PR to fix it!  It is very appreciated!

PR #1447, opened 2019-07-14
[Gtk] Always use headerbars

harry-cpp:
TLDR. AppMenu has been deprecated, and Gtk 3 apps should by general use Headerbars.

On a related note, I'll look into making toolbar be a headerbar for Gtk 3 apps.


harry-cpp:
@cwensley build bot down?

cwensley:
@cra0zy I've disabled the VS for Mac addin build for now so we can keep things moving along.  Thanks for the PR!

PR #1489, opened 2019-10-14
bind to NSSavePanel Message instead of Title

acormier:


acormier:
ignore. see #1490 instead

PR #1490, opened 2019-10-14
MacFileDialog: bind Title property to NSSavePanel Message because NSS…

acormier:
…avePanel Title isn't displayed anymore


cwensley:
Thanks @acormier, looks great!

PR #1494, opened 2019-10-23
SelectFolderDialog.Title wraps NSOpenPanel Message instead of Title

acormier:
Because macOS doesn't display Title anymore.


cwensley:
Thanks, @acormier!

PR #1519, opened 2019-11-20
WindowState.FullScreen support

cwensley:
Only implemented in macOS so far, but now we need to implement it for WinForms, WPF, and Gtk (if possible), and/or map it to Maximized.


harry-cpp:
> and Gtk (if possible)

https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-fullscreen

CrkwProject:
how is the state for this? seems an old pullrequest, is now implemented?
This is for giving the capability to trigger the fullscreen mode within the app only or pressing F11 fullscreen is not toggled too? 
Today, almost all the desktop/window managers allows to use F11 without the need of adding more code to your app, but if you want to toggle fullscreen from the app itself, you must to check the window manager (ie, in gtk you should track the window-state-event).

I don't think WindowState.Fullscreen is the place, WindowState should have minimized, normal or maximized, fullscreen would be a method to toggle the fullscreen, but fullscreen is just the maximized without titlebar and setting the form on top of everything, so can be easier to do that and maybe in windows changing the taskbar visibility to hidden while in fullscreen, but this is not the best aproach, if your app crashes and you don't handle the exception, you can't close it properly and return the taskbar visible again. Making changes like that requires a very controlled code and maybe having a service or daemon proccess different from the app itself.

PR #1523, opened 2019-11-23
[Gtk] Minnor fixups to NumericStepper and TextBox

harry-cpp:
Stuff done:
- Set `WidthChars ` to` 0`, this property is used as determining the minimum size for Gtk.Entry
- Removed Wrap, it was causing a bug, plus that is not how a Gtk stepper is supposed to act by default


cwensley:
Awesome, thanks!  I wondered why the NumericStepper was showing so large.  I assumed it was because it was using the max/min values to determine its width.

PR #1524, opened 2019-11-24
[Gtk] Fix setting smaller sizes for drop down not having effect

harry-cpp:
Yes, I intentionally added the sample drop down to have smaller width.


cwensley:
Beautiful, thanks again @cra0zy!

PR #1542, opened 2019-12-23
[GTK] Fix message box buttons not closing message box

MartinRothschink:
Close (hide) message box after click on any message box botton.


cwensley:
Thanks again!  perfect.

PR #1545, opened 2019-12-30
Fix Use WinForms and WPF tags

fredrikhr:
The test project for WinForms is missing the `UseWindowsForms` property whereas the WPF test project has one to many. This causes a warning during build. This PR fixes this issue.

### DIFF
--- a/test/Eto.Test.Wpf/Eto.Test.Wpf.csproj
+++ b/test/Eto.Test.Wpf/Eto.Test.Wpf.csproj
@@ -18,7 +18,6 @@
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <Prefer32Bit>false</Prefer32Bit>
     <UseWPF Condition="$(HaveWindowsDesktopSdk) == 'True'">true</UseWPF>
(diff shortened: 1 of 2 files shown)


test/Eto.Test.Wpf/Eto.Test.Wpf.csproj
```diff
@@ -18,7 +18,6 @@
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <Prefer32Bit>false</Prefer32Bit>
     <UseWPF Condition="$(HaveWindowsDesktopSdk) == 'True'">true</UseWPF>
-    <UseWindowsForms Condition="$(HaveWindowsDesktopSdk) == 'True'">true</UseWindowsForms>
```
cwensley, line 21: This appears to cause a build error, as Eto.Test.Wpf uses both WinForms and WPF for its unit tests.

cwensley:
Thanks @couven92 for the PR.  It appears to have build errors, though I'm not sure why it is only showing up in appveyor.  However, the Eto.Test.Wpf project uses both winforms and wpf so both of those properties are needed.  

Regardless, I'm working on some updates to support .NET Core with the Mac64 platform and I'll include this fix if you haven't updated your PR yet.

PR #1557, opened 2020-01-22
Use unified namespaces for MonoMac

cwensley:
For better compatibility with Xamarin.Mac

Not sure if I'll merge this just yet.. mostly just a test and to see if others are interested in such a change (please chime in either way).

This gets rid of the "MonoMac" prefix of the namespaces for better code compatibility with Xamarin.Mac, with much fewer `#if` blocks in the code.

This will, however, break any existing code/binaries that relies on the existing MonoMac namespaces, such as Eto.OpenTK, Eto.HtmlRenderer, etc. so they'd need to be updated to work with this.

### DIFF
--- a/src/Eto.Mac/AppDelegate.cs
+++ b/src/Eto.Mac/AppDelegate.cs
@@ -1,35 +1,17 @@
 using Eto.Forms;
 using System.ComponentModel;
 using Eto.Mac.Forms;
-
-#if XAMMAC2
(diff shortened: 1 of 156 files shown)


cwensley:
Closing, I don't want to break things, even in 3.x.  Eventually we'll all probably be using net6.0-macos with Xamarin.Mac instead, which is very promising.

PR #1573, opened 2020-02-07
Remove obsolete System.Xml reference for RoslyCodeTaskFactory

fredrikhr:
The `UpdatePList` task used in `Eto.Platform.Mac64.targets` uses a MSBuild `CodeTaskFactory`. Depending on whether we're on .NET Core or Full Framework we use either the `CodeTaskFactory` or `RoslynCodeTaskFactory` as indicated by our `$(MSBuildTaskFactoryType)` property.

The Code fragment in `Mac.targets` uses an XML Writer to generate a plist file for the Mac app bundle. On .NET Framework you'd reference `System.Xml` and load it from the GAC and this is what the previous version of `Mac.targets` does.

On .NET Core, `System.Xml` does not exist (and neither does the GAC). So the Xml types need to be referenced from somewhere else, namely the .NET Standard.

This PR creates a conditional MSBuild property `_UpdatePListReference` (following the MSBuild naming convention of internal properties being prefixed with `_`) that is evaluated either to `System.Xml` (on Full Framework) or `netstandard` (on .NET Core).

**Effectively, this makes it possible to use .NET Core compilers to produce Mac App bundles.**

### DIFF
--- a/src/Eto.Mac/build/Mac.targets
+++ b/src/Eto.Mac/build/Mac.targets
@@ -65,6 +65,13 @@
     <RemoveDir Directories="$(OutputAppPath)"/>
   </Target>
 
+  <PropertyGroup Condition="'$(MSBuildTaskFactoryType)'=='CodeTaskFactory'">
+      <_UpdatePListReference>System.Xml</_UpdatePListReference>
+  </PropertyGroup>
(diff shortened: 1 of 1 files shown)


cwensley:
Awesome, thanks for figuring this out and submitting the PR!  It's very appreciated.

PR #1574, opened 2020-02-08
Fix Eto Test project with VS Code

harry-cpp:


cwensley:
Awesome, thanks!

test/Eto.Test.Gtk/Eto.Test.Gtk.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
cwensley, line 5: This is actually incorrect.  It makes the Eto.Test.Gtk project _only_ build for .NET Core instead of both .NET Core and .NET Framework..

harry-cpp: Oh, than the solution is to edit the build script to use `msbuild` and not `dotnet build`.

cwensley: Yes, or we can specify the TargetFramework as part of the build options for dotnet, as I've done with #1580.  If you or anyone else cares to run/debug full .NET Framework versions using VSCode we can switch to using msbuild instead.

I think there's also a NuGet package that we can use to build full .NET Framework assemblies with dotnet, which might be interesting to look into.  It would be nice to eventually be "free" of mono being a requirement and only rely on the dotnet sdk.  There's a few other things to do but the list is getting (much) smaller.

PR #1598, opened 2020-02-25
Added RegexTestBench to the list of apps using Eto

CommonLoon102:


cwensley:
Awesome, great work!  It would be nice to see a macOS version too 😀

CommonLoon102:
If somebody has MacOS and tests the app on it and opens a PR with the instructions, I'll gladly merge it. I don't like Apple.

PR #1620, opened 2020-03-08
Fix for PropertyGrid that on GTK didn't show expandable item

halid-durakovic:
I had an issue with PropertyGrid control and searched what is wrong with my code. Then I have discovered that problem is here.


cwensley:
Hey @halid-durakovic, thank you for the fix!  Great work.

PR #1662, opened 2020-04-24
Fix ComboBoxCell drop down items not selectable in every case

### LINKED ISSUE #1658: WPF ComboBoxCell: only elements within the rows bounding box can be selected
While the issue is mostly resolved (it works generally - but not all values can be selected), exactly the described issue still persists. When trying to select an element outside the bounding rectangle of all rows - but within the gridview, the drop down will just close and no value will be selected.

To be clear, with "rows bounding box" i mean the data area and specifically not the `GridView`s bounding box. *That* works. Every element of the drop down in the marked red areas can be selected, only "Normal" does not work:

dropdown-bounding

## Expected Behavior
I can select all elements that are available in the drop down.

## Actual Behavior
Only those elements which lie within the bounding rectangle *or* outside the grid control of the rows can be successfully clicked on. Even the same element can't be selected if the mouse pointer is outside the bounding box, while it works within.

## Specifications

  - Version: 2.5.1-ci-20200407.2 (should be the latest)
  - Platform(s): WPF
  - Operating System(s): Windows 10

ManuelHu:
* Not all ComboBoxCell drop down items were selectable, when the item was outside the rows bounding box.
* The mouse click was treated like a normal click in the whitespace area below the rows, thus the SelectionChanged event of the ComboBox never got invoked.
* This discarded the information about just clicked-on drop down item, and the data was never updated

* By moving this whitespace-click handling code to a new event handler this was fixed.

Fixes #1658.

### DIFF
--- a/src/Eto.Wpf/Forms/Controls/GridHandler.cs
+++ b/src/Eto.Wpf/Forms/Controls/GridHandler.cs
@@ -148,6 +148,7 @@ namespace Eto.Wpf.Forms.Controls
 				GridLinesVisibility = swc.DataGridGridLinesVisibility.None,
 				Background = sw.SystemColors.WindowBrush
 			};
+			Control.MouseUp += HandleOutsideMouseUp;
 		}
 
 		protected ColumnCollection Columns { get; private set; }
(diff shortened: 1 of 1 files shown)


cwensley:
This is awesome, thanks @ManuelHu!  I'll look into your fix and merge if I find no issues.

cwensley:
Looks good, thanks again!

PR #1695, opened 2020-05-26
Hotfix for Application.Instance.Open(url) not working in .net core

piersdeseilligny:
This is a small hotfix for `Application.Instance.Open()` not working when fed a URL on WPF, WinForms, and GTK with .net core.

The bug results from a breaking change in .NET Core, where `UseShellExecute` is `false` by default, as opposed to `true` on .net framework ([source](https://github.com/dotnet/core/issues/4109#issuecomment-573307566)). This pull request explicitly sets it to true, ensuring that it'll work no matter the framework.


cwensley:
Hey @piersdeseilligny, awesome!  Thanks for the fix!

PR #1717, opened 2020-06-11
Localize file and help menu items on all platforms

### LINKED ISSUE #1716: Localizable default Menu titles on every platform (Especially Gtk, Wpf)
On the Eto.Mac{,64} and XamMac2 platforms all menu default items can be localized using the `Application.LocalizeString` event.

On all other platforms (Gtk, Wpf, WinForms) this is currently not the case; but I think that would be useful, not only for consistency between platforms.

If you want, I can prepare a pull request...

## Specifications

  - Version: Current CI build
  - Platform(s): Gtk, Wpf, WinForms
  - Operating System(s): n/a

ManuelHu:
Fixes #1716.

It compiles and I suppose it will work... But I currently cannot test WPF and WinForms.


cwensley:
Hey @ManuelHu, this is great!  Thanks so much for the PR!

PR #1773, opened 2020-08-16
Make sliders use floating point values

### LINKED ISSUE #1772: Sliders use integer values instead of floating point values
A large majority of platforms support floating point values, where the abstraction uses integers for the slider value. This severely limits the usefulness of sliders. 
This might be this way is because of one or more platforms only supporting integer values for sliders, however that really shouldn't limit every other platform just for that, it could instead just round to nearest integer.

```cs
// Shortened for brevity
namespace Eto.Forms
{
    [Handler(typeof(IHandler))]
    public class Slider : Control
    {
        public int MaxValue { get; set; }
        public int MinValue { get; set; }
        public int Value { get; set; }
    }
}
```

## Suggested change
```cs
// Shortened for brevity
namespace Eto.Forms
{
    [Handler(typeof(IHandler))]
    public class Slider : Control
    {
        public double MaxValue { get; set; }
        public double MinValue { get; set; }
        public double Value { get; set; }
    }
}
```

## Specifications

  - Version: 2.5.4 NuGet package

InfinityGhost:
This change makes the Slider control use floating point values for all platforms. It appears that only WinForms is the only platform that doesn't support floating point values on sliders (because its so primitive), which has been accommodated for. 

Closes #1772 

## Affected, tested, and functioning as expected
- [x] Gtk
- [ ] Gtk2
- [ ] Gtk3
- [ ] Mac
- [ ] WPF
- [ ] WinForms
- [ ] WinRT
- [ ] iOS

### DIFF
--- a/src/Eto.Gtk/Forms/Controls/SliderHandler.cs
+++ b/src/Eto.Gtk/Forms/Controls/SliderHandler.cs
@@ -5,9 +5,9 @@ namespace Eto.GtkSharp.Forms.Controls
 {
 	public class SliderHandler : GtkControl<Gtk.EventBox, Slider, Slider.ICallback>, Slider.IHandler
 	{
-		int min;
-		int max = 100;
-		int tick = 1;
+		double min;
+		double max = 100;
+		double tick = 1;
 		Gtk.Scale scale;
 
 		public SliderHandler()
@@ -64,7 +64,7 @@ namespace Eto.GtkSharp.Forms.Controls
 			}
 		}
 
-		public int MaxValue
+		public double MaxValue
 		{
 			get { return max; }
 			set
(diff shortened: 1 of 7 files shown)


cwensley:
Hi @InfinityGhost, thanks for the PR it is really appreciated!

While I agree it would be nice if the sliders where floating point vs. integers, this change is an ABI breaking change so it will have to wait for v3.0 to get merged in.

InfinityGhost:
Is there an ETA on 3.0?

cwensley:
@InfinityGhost sorry, no definitive timeline is set for 3.0.  Probably some time next year for its first official release as there's a lot of work to do.

CrkwProject:
Sliders and other controls of that kind in .net, can use the property maxiumvalue. 
In example, setting 1000 instead of 100, making a value as 102 would be like 10.2

PR #1795, opened 2020-10-06
Improve doc

sagarkava:


cwensley:
Hm, sorry I'm not sure why we need this comment in build.yml?  It isn't only used for pull requests, either..

PR #1800, opened 2020-10-15
Augmenteddrawloop

derbyw:
This augments the Draw Loop section of the test app to add displayed text coordinates to all the shapes. The purpose is to exercise/stress the raw text functions a bit more in the draw loop.  The mode has an on/off checkbox at the top of the display. 
As there is only 1 drawable control, leaks associated with this process will rise slower than if multiple drawable controls exist but any issues should still show a slow rise in process memory consumption over time (30 min+)

### DIFF
--- a/src/Eto.Gtk/Forms/Controls/DrawableHandler.cs
+++ b/src/Eto.Gtk/Forms/Controls/DrawableHandler.cs
@@ -96,11 +96,10 @@ namespace Eto.GtkSharp.Forms.Controls
 				{
 					if (h.SelectedBackgroundColor != null)
 						graphics.Clear(h.SelectedBackgroundColor.Value);
-
+					
 					h.Callback.OnPaint(h.Widget, new PaintEventArgs(graphics, rect.ToEto()));
 				}
 			}
-
 #endif
 		}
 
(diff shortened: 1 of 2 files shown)


src/Eto.Gtk/Forms/Controls/DrawableHandler.cs
```diff
@@ -91,14 +92,29 @@ public void HandleDrawn(object o, Gtk.DrawnArgs args)
...
 				{
 					if (h.SelectedBackgroundColor != null)
 						graphics.Clear(h.SelectedBackgroundColor.Value);
-					
-					h.Callback.OnPaint(h.Widget, new PaintEventArgs (graphics, rect.ToEto()));
+
+					h.Callback.OnPaint(h.Widget, new PaintEventArgs(graphics, rect.ToEto()));
+				}
+
+#else
```
cwensley: We could just remove the old section

cwensley:
Looks great, @derbyw!  Thanks for submitting the PR.  Please just remove the `#if old` section and it's good to go.

cwensley:
Ugh, editing via GitHub is a pain. Anyway, this is now merged with your previous changes.  Thanks again for submitting these PR's!

derbyw:
Sorry – didn’t see this till now – I see you took care of it already..

 

-Bill

Issue #1796, opened 2020-10-06
GTK Platform/GtkSharp has a slow memory leak with Drawable controls


derbyw:
I have an application that uses many Drawable controls to put up custom gauges for my application.
Updating these controls will slowly consume memory/resources on the Gtk platform running on GTK+3
on both Unix and Windows. The memory issue does not occur on the WPF build and I have no mac to test there.

To isolate the problem, I have created the following test app that demonstrates the issue and put up on GitHub:    

https://github.com/derbyw/EtoGtkReseourceLeakTest.git

see the readme for info on what buttons do. there are options to stop/start the updates

The app puts up a bunch of controls (16 text 4 bar gauge) and feeds them at about 100ms - the app will display a running total of memory which will slowly climb as the app runs.  If you let it run in VS 2019 you can also watch it rise over time - give it a good 10 minutes to see the distinct ramp.

The app is currently running against the current release (2.5.6) built in VS 2019. running against the msys2 Gtk+3 system on Windows and also shows against the Gtk+-3.22.26 on Linux (embedded) and also Ubuntu LTS 18.04.  (can't recall the GTS version there).

I had run into this previously on my original build for the embedded system and ended up fixing some of the auto generated code in GtkSharp after a bunch of time spent with Valgrind, etc. That system also supported the Gestures changes I proposed a while back and were not tied up cleanly to get back to Harry for GtkSharp.  So I've been stuck at Eto.Forms from about 2 year ago till I had enough time to fix things again and get sync'd with the project. I finally got the time, ported everything to .Net Core 3.1, crossing my fingers that the issue was fixed in the meantime -- alas the issue remains.

IIRC - the leak was be related to the various resources allocated and release when the drawable control get drawn i.e. each point event.  The reference counting seemed to get messed up and "freed" Pango, etc. resources don't quite all get returned. I will try to find my changes from when I fixed this last time to see what I modified. I was able to finally get the process net memory neutral and have been running that way for the last 2 years.  Unfortunately, I have not been able to get the new GtkSharp build to build on my embedded machine so I will continue work there.

derbyw:
for context, here are my 2017 notes from the fixes to the original:

 I have a GTK3 Eto.Application form with a bunch of custom (i.e. Drawable) based controls (text and bar gauges) and for the last several weeks have been chasing down a 
[...]

PR #1801, opened 2020-10-15
Fix slow leak of Pango context objects in drawable update

derbyw:
The change to use the Gtk# Pango (text) context associated with the control instead of creating a new local one on each draw event.  The leaked objects would be detected as a slowly growing process memory footprint the rate of which was determined by the amount of Drawable update activity. Tested on several (windows/Linux) platforms for a few hours with stable process memory.
addresses issue #1796


cwensley:
Beautiful!  Thanks for the fix.

PR #1803, opened 2020-10-16
Gestures 413 rev2

derbyw:
original functional GTK gestures and gesture event aware winforms support


cwensley:
Gestures were fully implemented on Mac, Gtk, Wpf, WinForms in #2950

Issue #1816, opened 2020-11-06
Eto.Gtk DateTimePicker drop-down dialog fails to close and loses parent connection.


derbyw:
I'm seeing an issue with the DateTimePicker implementation in Eto.Gtk where after the "drop down" arrow is clicked the child dialog window (with the calendar/clock controls on it) appears but can become stuck and will not disappear - it appears to becomes un-parented as you can move the parent form away and the picker will stay in place until you specifically click on the form. On my program, I can even close the parent window and the dialog will remain. On the Demo program this can be seen by clicking the drop down a few times in a row - on my system it works the first time, but eventually the dialog form no longer responds and just stays on the screen.

After more investigation, this appears to be a UI/Threading issue -- the "Close" command in CloseDialog for the DateComboBox gets called but doesn't work - if I manually close the form by managing the dlg reference to the form in DateComboBox.cs then the Close works as expected, but in that case, it is running on the UI thread -- is there a way to run something on the UI thread down in in BaseComboBox.gtk3.cs level of things? From what I can tell the Gtk Close() likely needs to run on the UI thread to work correctly and at the moment is running in an event handler thread.

derbyw:
Further investigation shows it's not threading (directly) - I added tried changing the code to capture a SynchronizationContext on the popup call and use that to call context Close -- to no effect.

It MAY be an issue with the Gestures as I am using the modified version to add that. I will switch to "develop" branch to see -- 

I do have a mod make it work - mostly - where the form always comes down but you sometimes have to click on the pull-down button twice to get it to appear -- also seems more likely to happen with fast click rate on the dropdown button.

derbyw:
Not related to gestures -- same issue happens on the "develop" branch.

derbyw:
I now have a usable version of this now but a few things need to be different - for one, the prior version of the dialog did a grab to allow a "click anywhere" to close the dialog - which unfortunately made is so the first click on the calendar would close  the dialog (so if you wanted to change year and date you needed to bring it up twice. 
After removing the grab you can now close the dialog by clicking on a non-control, re-clicking to drop down button or moving focus off the control. 
The one thing I'm still tying to get to work is for the dialog to either (a) track the entry field position if the parent form is moved or (b) close the dialog if the parent form is moved.  Note sure I will get very far with either and current is at least equal to the existing in that regard.  I'll create a PR for this so you can try the changes shortly.

cwensley:
[...]

PR #1818, opened 2020-11-09
#1816 fix parenting/close behavior for Gtk drop-down dialog box

derbyw:
This still needs some cleanup but the control no longer loses parenting for the drop-down on a form.  This result is still messed up if the control is located on a Dialog - guessing the two fight for the always on top but at least will behave on a Form. I have seen this problem in other Gtk situations and my response is to turn all my "dialogs" into forms so the only remaining are the "system" dialogs e.g. Save/Open/Print. Basically the code changes allow the combo-box part of the control to manage the lifetime of the dialog and will pull the dialog down when focus is lost to the control

As far as root cause -- the "Close" on the dialog gets called but seems to only partially close the window -- perhaps the underlying controls in Gtk-land abort the Close or there is a race condition or..???   I originally thought a race condition or non-UI thread since these are coming in from events - but I tested calling/sending the Close request to the context used to create the control to no effect -- seems something is "cancelling" the close request somehow.  I may move my investigation over to the UNIX side so I can debug into GtkSharp and see what's going on down below.

### DIFF
--- a/src/Eto.Gtk/CustomControls/BaseComboBox.gtk3.cs
+++ b/src/Eto.Gtk/CustomControls/BaseComboBox.gtk3.cs
@@ -24,7 +24,10 @@ namespace Eto.GtkSharp.CustomControls
 		{
 			if (args.Event.X > AllocatedWidth - ArrowWidth && Sensitive)
 			{
-				OnPopupButtonClicked(EventArgs.Empty);
+				if (args.Event.Type == Gdk.EventType.ButtonPress)
+				{
+					OnPopupButtonClicked(EventArgs.Empty);
+				}
 			}
 		}
 
--- a/src/Eto.Gtk/CustomControls/DateComboBox.cs
+++ b/src/Eto.Gtk/CustomControls/DateComboBox.cs
@@ -99,6 +99,8 @@ namespace Eto.GtkSharp.CustomControls
 				// if we are still up somehow, take the old dialog down..
 				if (dlg != null)
 				{
+					// this will appear to the user as a click that doesn't 
+					// bring up a window..
 					dlg.Close();
 					dlg = null;
 				}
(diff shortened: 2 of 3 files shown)


derbyw:
I see no good way at the widget level in Gtk to know when the form is being moved/resized. Perhaps a solution would be to simply add an "Expanded" property to the Eto DateTimeControl control to allow an app, if interested, to programmatically query, expand/collapse the dialog -- but then the support would need to be added in all the other modalities. Opinion?

cwensley:
Could we close the form when it loses focus?  Or does it not lose focus when moving the parent dialog?

derbyw:
Unfortunately, it doesn't - only the GtkWindow widget knows it's being moved and you can't really got to in in GtkSharp -- the "ParentWindow" gets you a GdkWindow which is the rectangle that the GtkWindow is in. In theory, you can get to that as part of the UserData but the whole things started feeling like a giant kludge.

The more I looked the worse it got - trying this out on a non-Ubuntu Wayland implementation (touch), the current "lose focus" thing breaks when the OSK pops up (which is does when you click on the Entry combo box button -- which then clobbers the drop-down. 

If you go back to the grab/click anywhere else approach to close, it's completely unusable on a touch screen (vs. mouse) because the targets are so small (even if I increase the font size).  I think part of the problem is that trying to create a drop-down dialog in Gtk depends on both GTK and the Linux compositors doing sensible things --- which they don't. 

Went through a lot of experiments and variations this morning to no avail. At this point I just implemented my own dialog box (in Eto) and threw the "calendar" and "clock" part overboard - they're pretty but not worth the time and trouble.

Probably the "real" portable solution here would be to implement this dialog in Eto and draw the calendar and clockface on the same form as the Entry (not a child window) 

The update performs better than the existing as long as you have a mouse and no OSK.

cwensley:
#1816

Issue #1865, opened 2021-01-05
WebView2 runtime not detected on 32-bit Windows


lubos:
As per [Microsoft docs](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution#deploying-the-evergreen-webview2-runtime), registry key for WebView2 runtime on 64-bit Windows is at:

`HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}`

and on 32-bit Windows it's just:

`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}`

Eto.Forms is looking at the location used by 64-bit Windows only. This means users on 32-bit Windows are getting installation notice to install WebView2 runtime even after installing it.

cwensley:
Ah, thanks for reporting the issue.  They actually had [documented it incorrectly](https://github.com/MicrosoftDocs/edge-developer/commit/f2dbbfcb49b468a74546fac22c2f7ffd8944d2ed) when I looked at it initially, but is now fixed.  I'll update it shortly.

cwensley:
Should now be fixed in latest develop branch and will be part of 2.5.10.  Thanks again!

PR #1872, opened 2021-01-08
WebView fixes

cwensley:
Wpf/WinForms: Detect WebView2 on 32-bit windows properly
Wpf/WinForms: Fix installation of WebView2 on .NET Framework
Mac: Fix warning when printing


cwensley:
Fixes #1865

PR #1981, opened 2021-07-09
Wpf: Fix CheckBoxCell not firing CellEdited events

Sejsel:
`CellEdited` events were not fired if clicking the CheckBox and going through the `OnMouseUp()` override that helps skip the three required clicks (select, enter edit, click) because that does not end up going through the typical WPF edit path (and thus CommitCellEdit is not called).


cwensley:
Awesome!  Thanks for the fix.

PR #2003, opened 2021-07-26
WebView2Handler: some properties from private to protected

acormier:


cwensley:
Looks great!  Thanks.

PR #2006, opened 2021-08-03
Set deletable text in RichTextArea correctly;

onepiecefreak3:
Set Text property of the valid text range to delete, to string.Empty instead of null.


cwensley:
Awesome!  Thanks for the fix!

PR #2055, opened 2021-11-14
Change LinuxTrayIndicatorHandler to use libayatana-appindicator3 instead of libappindicator3

sedme0:
The package that provides `libappindicator3.so.1` was [removed from Debian 11](https://www.debian.org/releases/bullseye/amd64/release-notes/ch-information.en.html#noteworthy-obsolete-packages). `libayatana-appindicator3.so.1` is a drop-in replacement for `libappindicator3.so.1` (as far as I can tell), so this should work, and as far as I've tested, it has. I'll admit I'm not very cognizant of other distros, so it might break compatibility there. All I know is that it works on my machine, and that's all I have time to test, so I encourage others to test it on other distros.

### DIFF
--- a/src/Eto.Gtk/Forms/LinuxTrayIndicatorHandler.cs
+++ b/src/Eto.Gtk/Forms/LinuxTrayIndicatorHandler.cs
@@ -11,7 +11,7 @@ namespace Eto.GtkSharp.Forms
 {
 	public class LinuxTrayIndicatorHandler : WidgetHandler<GLib.Object, TrayIndicator, TrayIndicator.ICallback>, TrayIndicator.IHandler
 	{
-		const string libappindicator = "libappindicator3.so.1";
+		const string libappindicator = "libayatana-appindicator3.so.1";
 		Image image;
 		string imagePath;
 		ContextMenu menu;


cwensley:
Hey @sedme0, thanks for the PR!  Do you think it's worth it to support both libraries?  Or is this new library also available on older linux versions?

Miepee:
I checked the Arch Repos on this, while `libappindicator-gtk3` has the description of `Allow applications to extend a menu via Ayatana indicators`, however, the actual library name is still `libappindicator3.so.1`.

Ubuntu has both `libappindicator3-1` as well as `libayatana-appindicator3-1`. Both are available from 18.04 up until current 21.10.

Opensuse only has `libappindicator3` which provides `libappindicator3.so.1`.

And the same goes for Fedora and Slackware.


So far, it's mostly just Debian who fell out of the line. They managed to [break](https://stackoverflow.com/questions/65978703/missing-libappindicator3-1-installing-slack) a [lot](https://github.com/electron/electron/issues/27527) of [applications](https://support.discord.com/hc/en-us/community/posts/1500000259801-unmet-dependencies-when-installing-on-debian11-bullseye-testing-) with this by [knowingly getting rid of appindicator](https://www.debian.org/releases/bullseye/amd64/release-notes/ch-information.en.html#noteworthy-obsolete-packages) entirely. So best way would probably just to support both libraries, since other distros don't plan on following suit with it.

cwensley:
Closing this as #2486 fixes the issue while also supporting the old versions.

PR #2098, opened 2021-12-31
Override incorrect dotnet paths in EtoSpecialFolders

### LINKED ISSUE #2096: GTK: EtoSpecialFolder.Downloads not following XDG user directories
## Expected Behavior
It should follow the XDG user directories on *nix systems and point to the user defined download directory.

## Actual Behavior
Currently, `GetFolderPath(EtoSpecialFolder.Downloads)` is hardcoded to basically return `$HOME/Downloads`.

On *nix systems, the user is able to set their own download folder, which is done frequently on localized systems (i.e. `~/Завантаження`).

## Specifications

  - Version: 2.6.0
  - Platform(s): GTK3
  - Operating System(s): Arch Linux
  
One can either get those by parsing the output of `xdg-user-dir DOWNLOAD` (which returns the full path) or by parsing the `$XDG_CONFIG_HOME/user-dirs.dirs` file.

Since GTK programs can be run on other OS', this would need an OS check to check if the current OS is on Linux. Windows has it hardcoded to `~/Downloads` iirc, not sure about MacOS.

Miepee:
- Use XDG_DOWNLOAD_DIR on Linux for EtoSpecialFolder.Downloads (Fix #2096) and the hardcoded download path for everything else.
- Use XDG_DOCUMENT_DIR on Linux for EtoSpecialFolder.Documents and the harcoded download folder for Mac
- Return XDG_CONFIG_HOME for EtoSpecialFolder.ApplicationSettings on Linux and HOME/Library/Preferences for Mac
- update some documentation

Note:
You *can* apparently rename core user folders in windows, and windows does update them somewhere in the registry. Thus, for EtoSpecialFolder.Downloads it would be more accurate to fetch the downloads folder from the registry.
The exact key is in `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders\{374DE290-123F-4565-9164-39C4925E467B}`, however there's also a really neat warning with `!Do not use this registry key -    Use the SHGetFolderPath or SHGetKnownFolderPath function instead` 🙃

### DIFF
--- a/src/Eto/EtoEnvironment.cs
+++ b/src/Eto/EtoEnvironment.cs
@@ -14,7 +14,7 @@ namespace Eto
 		/// <remarks>
 		/// This will return a different folder, depending on the platform: <br/>
 		///   OS X:    ~/Library/Preferences/ <br/>
-		///   Windows: %APPDATA% ([User's Home]/Appdata/Roaming <br/>
+		///   Windows: %APPDATA% [User's Home]/Appdata/Roaming <br/>
 		///   Linux:   ~/.config <br/>
 		/// </remarks>
 		ApplicationSettings,
@@ -41,10 +41,14 @@ namespace Eto
 		/// the location of the assembly can no longer be found as it is loaded from memory.
 		/// </remarks>
 		EntryExecutable,
-		
+
 		/// <summary>
 		/// Gets the user's downloads folder
 		/// </summary>
+		/// <remarks>
+		/// Note that for GTK on Windows, this will *always* return "[User's Home]/Downloads",
+		/// regardless of what the user's actual download path is.
+		/// </remarks>
 		Downloads
 	}
 
--- a/src/Eto/EtoEnvironment.cs
+++ b/src/Eto/EtoEnvironment.cs
@@ -12,10 +12,10 @@ namespace Eto
 		/// Application settings folder to store settings or data
 		/// </summary>
 		/// <remarks>
-		/// This will return a different folder, depending on the platform:
-		///   OS X:    ~/Library/Application Settings/[Name Of Application]
-		///   Windows: [User's Home]/AppSettings
-		///   Linux:   ~/.config
+		/// This will return a different folder, depending on the platform: <br/>
+		///   OS X:    ~/Library/Preferences/ <br/>
+		///   Windows: %APPDATA% ([User's Home]/Appdata/Roaming <br/>
+		///   Linux:   ~/.config <br/>
 		/// </remarks>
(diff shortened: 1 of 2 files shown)


ManuelHu:
While the currently returned folders are definitely not correct, I'm not sure if this change is great for a 2.x release - it will break all applications that assume e.g. a config file has been written in `EtoSpecialFolder.ApplicationSettings`.

Independent of the version this will be merged into, all those applications would need some some sort of (platform dependant) upgrade code that has to replicate Eto's old behaviour to be able to migrate old settings files etc. I'm not quite sure how this should be addressed...

Miepee:
Something that could probably help with this, is to not replace the original values.
Keep `EtoSpecialFolder.ApplicationSetting`, `EtoSpecialFolder.Documents` and `EtoSpecialFolder.Downloads` as they are but mark them as deprecated, and introduce something along the lines of  `ApplicationSettingsDirectory`, `DocumentsDirectory` and `DownloadsDirectory` which return the corrected values. That way devs can slowly migrate to the fixed variants with the old methods still available.

At the end of the day, it's up to @cwensley though to decide how this should be handled.

Miepee:
Took cwensley's suggestion in glitter/matrix and implemented a variable that determines now if the new folders should get returned or the legacy ones.
By default, it will return the new folders, but you can opt-out and use the old behaviour by setting `Eto.Gtk.EtoEnvironment.UseXDG` to `false` in your GTK application. The only affected variables by this are `ApplicationSettings` (Mac), `Documents` (Mac+Linux) and `Downloads` (Mac+Linux). Hope documentation for the variable is sufficient.

ManuelHu:
Yes, that solution sounds reasonable to me. 👍🏻

src/Eto/EtoEnvironment.cs
```diff
@@ -12,18 +12,18 @@ public enum EtoSpecialFolder
 		/// Application settings folder to store settings or data
 		/// </summary>
 		/// <remarks>
-		/// This will return a different folder, depending on the platform:
-		///   OS X:    ~/Library/Application Settings/[Name Of Application]
-		///   Windows: [User's Home]/AppSettings
-		///   Linux:   ~/.config
+		/// This will return a different folder, depending on the platform: <br/>
+		///   OS X:    ~/Library/Preferences/ <br/>
```
cwensley: This comment is now incorrect for macOS

cwensley:
Thanks @Miepee for the PR!  It looks great!

PR #2099, opened 2022-01-01
Readme: Add screenshots

Miepee:
Just adds some screenshots of Eto.Test to the readme.


cwensley:
Hey @Miepee, thanks for the PR!  I would have loved to see WPF screenshots instead of WinForms as I would prefer to push people to use that instead at this point (it has more features, has faster UI, and is better tested with Eto).  Perhaps soon I'll be creating a WinUI3 platform for Eto which would then be the preferred UI on Windows.  

Would you be able to switch out the WinForms screenshot with WPF?

Miepee:
Done!

cwensley:
Awesome, very much appreciated!!

PR #2154, opened 2022-03-01
Readme : Add 3'rd party NuGet/Library list

rafntor:
This adds a section listing available resources to the readme.

Or maybe a better place would be in the wiki? If you want this shameless promotion  at all ;-) 

preview : https://github.com/rafntor/Eto/tree/library_list#third-party-libraries


cwensley:
Love it!!!  thank you for putting these together.

PR #2169, opened 2022-03-23
update readme

rafntor:
include libvlc in library listing ;-)


cwensley:
Reminds me.. I should also add Eto.OpenTK, Eto.Veldrid, and Eto.HtmlRenderer to that list..

PR #2170, opened 2022-03-23
update readme ; add opentk, veldrid, htmlrenderer

rafntor:
any more you forgot ? :smirk:
.. or are aware of ?


cwensley:
Hm, oh right there's also Eto.CodeEditor.  It only was implemented for Wpf and XamMac2 so it isn't all that useful, but it could be updated if people are interested.

cwensley:
Hey @rafntor, thanks for adding these.  You have done some great work putting all of this together, making the Eto.Forms ecosystem much better.  I really appreciate it.

rafntor:
Thanks 👍

PR #2184, opened 2022-04-10
.NET 6.0.2xx SDK support

cwensley:
Looks like Xamarin.Mac.dll is now Microsoft.macOS.dll in .NET 6.0.2xx SDK with many breaking API changes so they are no longer compatible.

Eto.XamMac2 currently works with .NET 6.0.1xx SDK only.

This PR will add the `Eto.Platform.macOS` nuget package and fix all of the breaking changes.

Ideally the existing Eto.Platform.XamMac2 package will reference the above for .NET 6 going forward for a smooth transition between the two.


harry-cpp:
I was just about to ask you about this :D

Awesome stuff!

PR #2216, opened 2022-05-12
Wpf: Fix more cases where wrapped labels can cause endless update cycles

cwensley:


ievgen-baida:
Hello @cwensley, it seems this is the issue that causes high CPU usage on idle in our application. 
Do you have the timeline for releasing 2.7.1 with this fix?

PR #2236, opened 2022-06-03
Do not add child window to parent when the child window is in the TabbedWindows of parent

AkiSakurai:
There is a setting in macos system preferences to group windows as tabs automatically. And tabbed windows is not compatible to AddChildWindow relationship.


cwensley:
afaik, you can certainly have child windows that are also themselves tabs.  Do you have some code to reproduce the issue this is supposed to solve?
Tabbed Child Windows

AkiSakurai:
``` c#
using System;
using Eto.Forms;
using Eto.Drawing;

namespace Tutorial1
{
	public class MyForm : Form
	{
		static bool firstShown = true;

		public MyForm(Window parent = null)
		{

			if(parent != null)
			{
				this.Owner = parent;
			}

			if(firstShown)
			{
				firstShown = false;
				Application.Instance.InvokeAsync(() => (new MyForm(this)).Show());
				
			}
			// Set ClientSize instead of Size, as each platform has different window border sizes
			ClientSize = new Size(600, 400);

			// Title to show in the title bar
			Title = "Hello, Eto.Forms";

			// Content of the form
			Content = new Label { Text = "Some content", VerticalAlignment = VerticalAlignment.Center, TextAlignment = TextAlignment.Center };
		}
	}

	class Program
	{
		[STAThread]
		public static void Main(string[] args)
		{
			new Application().Run(new MyForm());
		}
	}
}

using System;
using Eto.Forms;
using Eto.Drawing;

namespace Tutorial1
{
	public class MyForm : Form
	{
		static bool firstShown = true;

		public MyForm(Window parent = null)
		{

			if(parent != null)
			{
				this.Owner = parent;
			}

			if(firstShown)
			{
				firstShown = false;
				Application.Instance.InvokeAsync(() => (new MyForm(this)).Show());
				
			}
			// Set ClientSize instead of Size, as each platform has different window border sizes
			ClientSize = new Size(600, 400);

			// Title to show in the title bar
			Title = "Hello, Eto.Forms";

			// Content of the form
			Content = new Label { Text = "Some content", VerticalAlignment = VerticalAlignment.Center, TextAlignment = TextAlignment.Center };
		}
	}

	class Program
	{
		[STAThread]
		public static void Main(string[] args)
		{
			new Application().Run(new MyForm());
		}
	}
}

```
If you click on the parent tab, the tabbar is gone, click again the whole window is gone


https://user-images.githubusercontent.com/75532970/171969683-5ac3162b-37d2-4f80-be3c-c7af55600256.mp4

AkiSakurai:
Also, I just tested your test program, it fails too. 
Maybe it is bug introduced by new macos version?
My os version is monterey 12.4.

AkiSakurai:
Just tested on high sierra, the issue exists.

cwensley:
Hm, yeah you're right.  Clicking on the "parent" window causes them to disappear, but then switching to another app then back brings them back.  I'll have to investigate why that's happening, you should certainly be able to have a child window even when using tabs.

cwensley:
Sorry this took so long to get in!  Thanks for the contribution!

PR #2286, opened 2022-08-21
Mac: Add a warning when the specified icon file is missing

cyanfish:
I found it confusing how the bundle always had the default icon no matter what file name I specified (I was missing the Content definition). Hopefully this prevents someone else from having the same problem.


cwensley:
Awesome! Thanks for submitting the PR.  This should certainly help other devs in the same situation.

Issue #2295, opened 2022-08-28
Wpf crashes on Windows 7


hamarb123:
## Expected Behavior
App runs and works on Windows 7.

## Actual Behavior
App crashes on Windows 7 (Wpf).

## Stack trace
![image](https://user-images.githubusercontent.com/25316035/187064489-752598f8-c92f-4fbe-a4be-fc03c4d7a5b8.png)

## Cause
The issue is that some apis, like this one, aren't supported on older versions of Windows, I'm happy to go through the code, find these, and guard them.

## Specifications

  - Version: 2.7.1 (was using 2.6.0 previously, edit: it seems to be fine in 2.7.0)
  - Platform(s): WPF
  - Operating System(s): Windows 7

cwensley:
Hm, I don't support Windows 7 as neither has MS (since January 2020).  However I will accept any fixes. 😄

hamarb123:
Fair enough, I would like to say that both .NET 6 and .NET framework 4.8 do support Win 7 though, although the upcoming newer versions will not 'support' it (but .NET 7 will not go out of their way to break on it). So I think was a reasonable assumption that it would support Windows 7, but I'm more than happy to fix any issues I encounter until I decide to not support Windows 7 anymore lol.

PR #2297, opened 2022-08-28
DPI API Fixes for Wpf on Windows 7

hamarb123:
This commit fixes the crashes I experienced as described in https://github.com/picoe/Eto/issues/2295. It makes everything work in my application and it should probably make everything else work too, but imo these APIs are not exposed or checked very well.

@cwensley I'd like to first merge this and then implement a much more thorough fix that will expose only APIs with error checking and fallbacks (i.e. 1 method for get dpi, try set awareness, get awareness, & whatever else is needed) - this would avoid something like this happening in the future with these APIs.

This PR just adds fallbacks to the `SystemDpi` and `IsSystemDpiAware` APIs that work on Windows 7.


cwensley:
Thanks for the fix!  Looks great.

PR #2299, opened 2022-08-29
Added DropDownToolItem.

allsorts46:
My implementation of DropDownToolItem, to compare/combine with #2243 from @Serg-Norseman. Also added example to the Eto.Test project.

Implemented and tested with Winforms, WPF and GTK. Android implementation also exists but depends on #2277.

### DIFF
--- a/src/Eto.Gtk/Forms/ToolBar/DropDownToolItemHandler.cs
+++ b/src/Eto.Gtk/Forms/ToolBar/DropDownToolItemHandler.cs
@@ -1,0 +1,90 @@
+using System;
+using Eto.Forms;
+using Gdk;
+
+namespace Eto.GtkSharp.Forms.ToolBar
+{
+
+	public class DropDownToolItemHandler : ToolItemHandler<Gtk.ToolButton, DropDownToolItem>, DropDownToolItem.IHandler
+	{
+		// GTK3 MenuToolButton looks kind of horrible with a huge visually separate drop button, and forces
+		// the main button and drop button to function separately. So, use a normal ToolButton with a Menu, but
+		// add a drop arrow to the button text.
+
+		private Gtk.Menu dropMenu;
+		
+		public DropDownToolItemHandler()
+		{
+			dropMenu = new Gtk.Menu();
+		}
+		
+		#region IToolBarButton Members
+
+		
+		#endregion
+
+		public override void CreateControl(ToolBarHandler handler, int index)
+		{
+			Gtk.Toolbar tb = handler.Control;
+
+			Control = new Gtk.ToolButton(GtkImage, Text + "  ▾");
+			Control.IsImportant = true;
+			Control.Sensitive = Enabled;
+			Control.TooltipText = this.ToolTip;
+			Control.ShowAll();
+			Control.NoShowAll = true;
+			Control.Visible = Visible;
+			//control.CanFocus = false;			// why is this disabled and not true???
+			tb.Insert(Control, index);
+			Control.Clicked += HandleClicked;
+		}
+
+		private void HandleClicked(object sender, EventArgs e)
+		{
+			dropMenu.PopupAtWidget(Control, Gravity.SouthWest, Gravity.NorthWest, null);
+			Connector.HandleClicked(sender, e);
+		}
+
+		protected new DropDownToolItemConnector Connector { get { return (DropDownToolItemConnector)base.Connector; } }
+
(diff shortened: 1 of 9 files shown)


allsorts46:
I encountered an issue in Eto.Test when trying the ToolBarSection on Winforms platform, but it looks like a problem with NumericStepper (or usage thereof) and so I did not commit my workaround on this branch.

I had to modify the SetStepperLimit() method from:

```
void SetStepperLimit()
{
	if (toolBar.Items.Count == 0)
		indexStepper.MinValue = indexStepper.MaxValue = -1;
	else
	{
		indexStepper.MinValue = 0;
		indexStepper.MaxValue = toolBar.Items.Count - 1;
	}
}
```

to:

```
void SetStepperLimit()
{
	if (toolBar.Items.Count == 0)
	{
		indexStepper.MinValue = -1;
		indexStepper.Value = -1;
		indexStepper.MaxValue = -1;
	}
	else
	{
		indexStepper.MaxValue = toolBar.Items.Count - 1;
		indexStepper.MinValue = 0;
	}
}
```

in order to avoid exceptions about invalid values at runtime. Was something about NumericStepper changed recently which broke this code?

allsorts46:
Switched to using ToggleToolButton which allows a 'pressed' look whilst the menu is open. Seems to handle all methods of closing the menu (clicking item, clicking outside, pressing ESC) correctly and looks good with default theme. Unsure if it could look weird with custom GTK themes though.

cwensley:
Awesome, this looks great!  I'll try to add the Mac version and then we can get this merged in.  Thanks for taking the time to submit the PR.

Serg-Norseman:
👍 👍 👍

allsorts46:
Thanks! I also have a LabelToolItem that I wanted to submit in future - basically just non-interactive text in the toolbar. Assuming that's something you'd approve of, I'd need your help for the Mac implementation on that as well. Would it make sense to do both at the same time?

Serg-Norseman:
> Thanks! I also have a LabelToolItem that I wanted to submit in future - basically just non-interactive text in the toolbar. Assuming that's something you'd approve of, I'd need your help for the Mac implementation on that as well. Would it make sense to do both at the same time?

Also a useful item!

Serg-Norseman:
Does the arrow turn off for you only in the Gtk implementation?

allsorts46:
> Does the arrow turn off for you only in the Gtk implementation?

Yes, I only added it on the GTK implementation... but I suppose it should be easy to make the option in a platform-independent way instead as a property on Eto.Forms.DropDownToolItem. For GTK, WPF and Android we are manually _adding_ the arrow anyway so can just choose not to, and the Winforms System.Windows.Forms.ToolStripDropDownButton has a ShowDropDownArrow property. Don't know about Mac.

Serg-Norseman:
In my program in several places such buttons have an arrow, in several they should not have. And it is desirable that the view be the same on all platforms. How can I do this without a shared property in Eto.Forms.DropDownToolItem?

allsorts46:
Winforms and WPF also now allow hiding the drop arrow. I stopped short of added a property on the cross-platform assembly because I don't know if Mac will be able to support it or not. If it turns out it can, we can just add it as a single property in the main Eto assembly instead, which will be much easier to use.

So assuming you have project references to the appropriate platform assemblies, all of these now work:

```
Eto.Style.Add<Eto.GtkSharp.Forms.ToolBar.DropDownToolItemHandler>("NoArrow", h => h.ShowDropArrow = false);
Eto.Style.Add<Eto.WinForms.Forms.ToolBar.DropDownToolItemHandler>("NoArrow", h => h.ShowDropArrow = false);
Eto.Style.Add<Eto.Wpf.Forms.ToolBar.DropDownToolItemHandler>("NoArrow", h => h.ShowDropArrow = false);
```

and then any DropDownToolItem with style 'NoArrow' will not have an arrow on GTK, Winforms or WPF.

cwensley:
Ok so I've worked on this a bit on Mac and I've got this almost ready to go.  ShowDropArrow will also be promoted to a property in DropDownToolItem, which I think makes sense if all platforms support it.

cwensley:
I've now merged a fix that include the DropDownToolItem based this implementation.  Please give it a go and let me know if there's anything amiss.  Thanks @allsorts46 and @Serg-Norseman for doing the legwork to get this implemented!  It will be a very useful feature for Eto.

PR #2307, opened 2022-09-12
Move from the license-challenged Xceed library to an alternative

philstopford:
License-aside, this fork (pre license change) also has compatibility with modern .NET releases, unlike the 3.6 version from Xceed.


cwensley:
Hey @philstopford, thanks for submitting the PR!  I've been meaning to look into this for a while and it's cool to see it doesn't take a ton of changes to get this to work.  

However, I need this to support both (at least during compilation) via a property or some other mechanism.   Another way might be to have a separate nuget package that targets the new toolkit so developers can choose, but I'm not sure I'm a fan of that approach.

The main reason is this is a big breaking change for any project that created custom controls relying on Xceed.Wpf.Toolkit, which is also why I have been hesitant to make this change in v2.x.  

It would be interesting to hear from others to see if this breaking change is a big deal or not.  It certainly is for the scenarios I work on with Eto.

philstopford:
I tend to use Eto from nugets, so having a variant that drops Xceed would be helpful.

philstopford:
Just wondering if there were more thoughts about this

cwensley:
Hey @philstopford, sorry for the late response to this one.  While I'd like to support using the DotNetProjects version, I still need this to support the other version so I can't merge this in unless there is that option.   I'm not sure what exactly I'd like to see on the nuget side e.g. having two WPF platforms sounds horrible.

What I'd really love would be to remove all uses of the toolkit entirely.  I believe is only used for `Stepper`, `NumericStepper`, `TextStepper`, `DateTimePicker`, `ColorPicker`, `ColorDialog`, `TextBox.PlaceholderText`.  Not a ton to replace but definitely will need work.

philstopford:
Closing as unlikely to be merged

richard-sim:
As a new user of Eto, I'd very much like to see this merged as I get warnings every time I build the WPF project - and critically, the warning is treated as an error by VSCode. Being a newbie to Eto, this seems like an error that'd break my application on some machines ("Works on my machine!"), but given this thread I take it that it's... just normal? There's a notable lack of documentation around this warning being expected behaviour, and its implications (if any). The only option to fix this currently appears to be to pull @philstopford's fork, build that, and build my application against that... again, as a newbie that isn't what I should be doing.

With command line builds:
`<snip>.Wpf.csproj : warning NU1701: Package 'Extended.Wpf.Toolkit 3.6.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .
NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' in
stead of the project target framework 'net6.0-windows7.0'. This package may not be fully compatible with your project.`

Loading the project in VSCode:
`[warn]: OmniSharp.MSBuild.ProjectLoader
        Package 'Extended.Wpf.Toolkit 3.6.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' instead of the project target framework 'net6.0-windows7.0'. This package may not be fully compatible with your project.`

`<snip>.Wpf.csproj(0,0): Error: Package 'Extended.Wpf.Toolkit 3.6.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' instead of the project target framework 'net6.0-windows7.0'. This package may not be fully compatible with your project.`

cyanfish:
In case anyone else comes across this, you can suppress the warning by adding a reference to your project with NoWarn:
`<PackageReference Include="Extended.Wpf.Toolkit" Version="3.6.0" NoWarn="NU1701" />`

Issue #2326, opened 2022-09-29
GTK and WinForms Label with WrapMode.None should not enable SingleLineMode


LeeARichardson:
Hi,
On GTK and WinForms, but not WPF (I can't check Mac) setting a Label to WrapMode.None also enables single line mode. There's no way to disable wrap but still allow multiple lines.
This is in contradiction to the documentation which states "Regardless of the mode, you can always add hard wraps by inserting newline characters."
https://github.com/picoe/Eto/blob/86f8ba20544b11447f30794718f3170ca971a726/src/Eto/Forms/Controls/Label.cs#L30
https://github.com/picoe/Eto/blob/86f8ba20544b11447f30794718f3170ca971a726/src/Eto.Gtk/Forms/Controls/LabelHandler.cs#L155
https://github.com/picoe/Eto/blob/86f8ba20544b11447f30794718f3170ca971a726/src/Eto.WinForms/Forms/Controls/LabelHandler.cs#L198

## Expected Behavior
Label with WrapMode.None should not restrict the text to a single line.

## Actual Behavior
Text with new lines is all shown on the same line due to the SingleLineMode being enabled.

## Steps to Reproduce the Problem

1. Set a Label's Wrap to WrapMode.None .

## Code that Demonstrates the Problem
```fsharp
let label = new Label()
label.Wrap <- WrapMode.None
```

## Specifications

  - Version: 2.7.1
  - Platform(s): Gtk3, WinForms
  - Operating System(s): Windows 10, Linux

cwensley:
Fixed with #2327

PR #2327, opened 2022-09-30
Allow newlines in unwrapped Labels

LeeARichardson:
See: https://github.com/picoe/Eto/issues/2326

### DIFF
--- a/test/Eto.Test/Sections/Layouts/TableLayoutSection/ScalingSection.cs
+++ b/test/Eto.Test/Sections/Layouts/TableLayoutSection/ScalingSection.cs
@@ -50,7 +50,7 @@ namespace Eto.Test.Sections.Layouts.TableLayoutSection
 			tableLayout.SetColumnScale(2);
 			tableLayout.SetRowScale(0);
 			tableLayout.SetRowScale(2);
-			tableLayout.Add(new Label { Text = "2x2, should not scale and be centered", BackgroundColor = Colors.Red }, 1, 1);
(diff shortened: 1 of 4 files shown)


LeeARichardson:
This would be a breaking change so would you be interested if I do a separate PR to add an additional WrapMode for 2.x? `WrapMode.NoneMultiline` maybe?
I'm not able to test other platforms so is there any other work required?

test/Eto.Test/Sections/Layouts/TableLayoutSection/ScalingSection.cs
```diff
@@ -50,7 +50,7 @@ public ScalingSection()
 			tableLayout.SetColumnScale(2);
 			tableLayout.SetRowScale(0);
 			tableLayout.SetRowScale(2);
-			tableLayout.Add(new Label { Text = "2x2, should not scale and be centered", BackgroundColor = Colors.Red }, 1, 1);
+			tableLayout.Add(new Label { Text = "3x3, should not scale and be centered", BackgroundColor = Colors.Red }, 1, 1);
```
cwensley, line 53: ?  Why was this changed?

cwensley:
Hey @LeeARichardson, thanks for submitting the PR!  Yes all platforms should behave the same, so if doing this brings Gtk and WinForms in line with Mac and WPF I'm all for it.

Adding a separate `WrapMode.Singleline` or something similar so one could set the label to be single line even though there are newlines might be useful, but it is equally as easy just to remove the newlines from the string.

cwensley:
Just did a test and everything looks great.  

Thanks for the contribution! 😊

PR #2348, opened 2022-11-02
Add notice regarding Wayland Locations

Miepee:
This adds a quick notice to both `Location`s, detailing that the attribute has basically no use on Wayland.


cwensley:
So is there no way to actually get/set window locations on Wayland?  I did some minor googling around and it appears this is a "feature" of Wayland, but want to be sure.

Miepee:
It indeed is not possible at all on Wayland. An intended design decision, partly for security, and probably partly due to simplicity too.

cwensley:
Well that's just weird. 🤷🏻‍♂️  At any rate, thanks for adding the notes!  It's much appreciated.

PR #2349, opened 2022-11-02
Use native file pickers for GTK

### LINKED ISSUE #2138: [Feature Request] Support XDG-Desktop-Portals for consistent file pickers
On Linux, it's possible to pass the environment variable `GTK_USE_PORTAL=1` to GTK applications in order for those to use XDG-Desktop-Portals for consistent file pickers and similar.
Currently, Eto does not support those. Here's a demonstration showing an Eto application trying to invoke a file picker and Firefox (a gtk program) run with the aforementioned environment variable:
image
image

As far as I can see, GTK already handles [invoking native file pickers](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkfilechoosernativeportal.c#L445) if the environment variable is set, which GTKSharp also [seems to support/call](https://github.com/GtkSharp/GtkSharp/blob/a23d44fb82975711cfcaa400cd7f3e4988d15bbf/Source/Libs/GtkSharp/NativeDialog.cs).

Miepee:
This uses the native GTK file pickers instead of the normal ones, which will result in the application choosing the system's file picker when the environment variable `GTK_USE_PORTAL` (or `GDK_DEBUG=portals` in the future) is passed on.
I have not tested this on either Windows or MacOS. This *should* lead to the picker using the native controls there as well, but cannot say for certain as I haven't tested it.
Fixes #2138 
Before:
grafik
After:
grafik

# Caveat
For some odd reason that I still haven't figured out, passing that environment variable also causes the application to choose some non-native GTK theme on Linux. Seems to only happen on Wayland.
Without the env var:
grafik
With the env var:
grafik

I am unsure what causes it exactly, as it only appears when the environment variable is passed on. Potentially a GTK# issue, but am completely clueless there.

### DIFF
--- a/src/Eto.Gtk/Eto.Gtk.csproj
+++ b/src/Eto.Gtk/Eto.Gtk.csproj
@@ -23,7 +23,7 @@ When building for the full framework, .NET Framework 4.6.1 or mono framework 5.1
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="GtkSharp" Version="3.24.24.34" />
+    <PackageReference Include="GtkSharp" Version="3.24.24.38" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Eto\Eto.csproj" />
--- a/src/Eto.Gtk/Forms/Controls/FilePickerHandler.cs
+++ b/src/Eto.Gtk/Forms/Controls/FilePickerHandler.cs
@@ -59,7 +59,13 @@ namespace Eto.GtkSharp.Forms.Controls
 
 		private void Savebutton_Clicked(object sender, EventArgs e)
 		{
+#if GTKCORE
 			var savedialog = new Gtk.FileChooserNative(string.Empty, null, Gtk.FileChooserAction.Save, null, null);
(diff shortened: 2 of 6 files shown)


Miepee:
Marking as draft until I can figure out why the function does not exist on GTK3 / GTK2, but does on GTKCORE

Miepee:
Actually, @cwensley would fixing that even be necessary / wished for considering GTK3 and GTK2 are deprecated?

cwensley:
@Miepee Yeah I need to spend the time to remove those from the CI builds, as they really shouldn't need to be built anymore. Feel free to #ifdef it out in the meantime.

Miepee:
Ifdef'ed it out for older GTK versions.
@harry-cpp you're a bit more familiar with GTK than I am, you don't happen to know why passing the environment variable on Wayland causes GTK to display fonts blurry and to pick the dark Adwaita theme? I can overwrite it manually with `GTK_THEME`, but the blurryness still remains.

cwensley:
Awesome, looks great!  Thanks for getting this put together!

Issue #2353, opened 2022-11-19
Eto.Platform.Windows not supporting fx461


Beyley:
## Expected Behavior
Eto.Platform.Windows should be available on all frameworks that net standard 2 is available on (fx461+ and coreapp3.1+)

## Actual Behavior
Eto.Platform.Windows is only available on fx462+ and coreapp3.1+

## Steps to Reproduce the Problem

  1. Make fx461 app
  2. Import through nuget
  3. Try to build

## Specifications

  - Version: 2.7.2
  - Platform(s): WinForms
  - Operating System(s): Windows 10
  
Unless there is a specific reason fx461 cant be supported, it should work under fx461, same with the rest of the Eto nugets (which are netstandard2), it should be as simple as bumping the framework version from fx462 -> fx461

cwensley:
Sorry, but .NET 4.6.1 is end of life, so it is not supported: https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

PR #2354, opened 2022-11-19
Bump Eto.WinForms support down to 461

Beyley:
Seems to just compile OOTB, and means Eto.WinForms can be used on all platforms the rest of the Eto projects can be used on

### DIFF
--- a/src/Eto.WinForms/Eto.WinForms.csproj
+++ b/src/Eto.WinForms/Eto.WinForms.csproj
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <HaveWindowsDesktopSdk Condition="$(HaveWindowsDesktopSdk) == '' and $(OS) == 'Windows_NT'">true</HaveWindowsDesktopSdk>
     
-    <TargetFrameworks>netcoreapp3.1;net462</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


Beyley:
Closes #2353

cwensley:
Not sure if you saw the comment on the other issue, but .NET 4.6.1 is no longer supported.  Is there a particular reason you need this?

Beyley:
> Not sure if you saw the comment on the other issue, but .NET 4.6.1 is no longer supported.  Is there a particular reason you need this?

I aw writing an app for an old vista cash register to launch/manage my games and such, as the machine is great for early 2000s games, and due to the age of the games, they can be fiddly and writing a launcher would help smooth that out

cwensley:
Hm, with [this chart](https://learn.microsoft.com/en-us/dotnet/framework/get-started/system-requirements), it says vista only supports up to .NET 4.6, which does not support .NET Standard 2.0 either.

Beyley:
> Hm, with [this chart](https://learn.microsoft.com/en-us/dotnet/framework/get-started/system-requirements), it says vista only supports up to .NET 4.6, which does not support .NET Standard 2.0 either. 

4.6.1 is supported on vista, just not advertised, it installs and runs fine, 4.6.2 is where it stopped working on vista

cwensley:
Interesting.. well I suppose there's no harm.  Thanks for the PR!

PR #2357, opened 2022-11-25
Remove Gtk2 and Gtk3 platforms

### LINKED ISSUE #2109: Remove Gtk2 platform
The Gtk2 platform has been deprecated since 2.5.0, and will be removed in a future 2.6 update.  This will allow larger updates to fix things like text/background colours and styling without major refactoring.

cwensley:
- now only one Gtk platform for linux support

I'm leaving the projects in for now in case anyone needs to compile these separately.  They will be removed in 2.7.4+ if no one chimes in.

Fixes #2109


dsoronda:
Why remove Gtk3 platform ? Or I misunderstand title ? GtkSharp  support for Gtk4 is not out yet. 
So, Eto.Forms will not support Gtk / Linux anymore ?

cwensley:
@dsoronda Yes it is a little confusing, but Gtk support will continue with Eto.  The "Gtk3" platform was for using the [gtk-sharp3](https://packages.ubuntu.com/search?keywords=gtk-sharp3) packages on linux only, which is not maintained, outdated, and only works with mono.  Eto's "Gtk" platform will continue to be supported and uses [GtkSharp](https://github.com/gtksharp/gtksharp) to target Gtk+3.

PR #2362, opened 2022-12-05
Unsubscribe from the input event handler

Dmitri-Kom:
Class: `Binding`, method: `RemovePropertyEvent`.
If we don't do that, `dotMemory `(and probably other 
memory profiling tools) will report about an event
leakage of `PropertyNotifyHelper`.
At the moment, nothing is done with the
EventHandler<EventArgs>, that is passed as an
input parameter to the `RemovePropertyEvent` method. 
Performing the change fixes the report of the memory
profiling tool.


cwensley:
Hey @Dmitri-Kom, thanks so much for the PR!  This looks great, thanks for catching that leak.

PR #2366, opened 2022-12-08
Fix grammar error in string->filefilter operator exception

Miepee:
In the exception text, it opened with a `'`, and never closed it.


cwensley:
Thanks for the PR! 👍

PR #2381, opened 2023-01-21
Return Ok on Gtk.ResponseType.Acept instead of Ignore

Miepee:
This should fix the GTKNative dialog boxes returning Ignore when they should have returned Ok


cwensley:
Awesome, thank you! 🙏

PR #2426, opened 2023-02-17
Bump Eto C# lang to 10, use file scoped namespaces

Miepee:
This bumps the C# lang version for the Eto project to 8, and converts every file that has a use for it to use file scoped namespaces.


Miepee:
Seems that I was mistaken, and turns out one can't actually use file scoped namespaces in .net framework. Oh well.

Miepee:
Reopened because apparently file scoped namespaces *can* be used with .net framework, as long as its compiled with a new-enough compiler

cwensley:
This is great!  Thanks!

PR #2433, opened 2023-03-04
Create an interface for controls to show that they can hide the border

Miepee:
This PR creates a new interface (`IHideControlBorder`) which defines whether controls can show/hide the controls around their borders.
This makes it possible to later get rid of duplicated documentation text, and also make it easier to figure out whether (custom) controls support that feature.
For example, when one wants to disable the borders for all possible controls that are in a grid view, one could just do something akin to
```cs
foreach (var control in GetGridViewItems())
{
    if ((control as IHideControlBorder) is not null)
        (control as IHideControlBorder).ShowBorder = false;
}
```

### DIFF
--- a/src/Eto/Forms/Controls/DateTimePicker.cs
+++ b/src/Eto/Forms/Controls/DateTimePicker.cs
@@ -28,7 +28,7 @@ public enum DateTimePickerMode
 /// Date/time picker control to enter a date and/or time value
 /// </summary>
 [Handler(typeof(DateTimePicker.IHandler))]
-public class DateTimePicker : CommonControl
+public class DateTimePicker : CommonControl, IHideControlBorder
 {
 	new IHandler Handler { get { return (IHandler)base.Handler; } }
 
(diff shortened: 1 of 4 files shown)


cwensley:
Hey @Miepee, I'm not really a fan of adding interfaces for common properties on controls.  Is this really necessary?  This adds a lot of code smell in my opinion.  There are also other common properties/functionality that could have specific interfaces.. I'm not sure where it would end nor do I want to see a bunch of these added.

Miepee:
My main concern was, that it's currently a bit difficult to know for which controls to disable `ShowBorders`. I guess a slower workaround would be to just use reflection.
Also, and this may just be me, but I personally find having copypasted documentation harder to maintain, because you don't always remember that a specific text was also used 5 times somewhere else.

Making an interface for a single control is probably a bad idea tho. If you've got any alternative ideas, I'd be open to them.

cwensley:
Closing this as I don't think adding interfaces for controls with common properties is a thing we want to implement.  I really appreciate the thought and effort that went into this. I'm open to other ideas but right now this isn't where I want to go with the framework.

PR #2434, opened 2023-03-04
Mark some tests as manual, fix ColorParse tests with invalid point

Miepee:
- Mark some tests that need manual intervention as manual (printer test and `RunIterationShouldAllowBlocking`)
- Fix a bug in `Color.TryParse` returning invalid Colors when the user is on a culture that uses something other than a `.` for decimal point (i.e. German culture with `,` trying to parse `rgba(12,34,56,0.5)`) and adds it to the test suite in order for it to not happen again.


cwensley:
Looks great!  Thanks for the fixes.

PR #2436, opened 2023-03-04
[Mac] Fix loading local resources on Webview

### LINKED ISSUE #2380: Loading local resources (JS, CSS, images) in WebView doesn't work on Mac when calling LoadHtml()
_Originally posted by @somelinguist in https://github.com/picoe/Eto/discussions/2379#discussioncomment-4709766_
          

This is an issue with WkWebView's security. If you search for trying to load local resources with webview.LoadHtmlString(), there are tons of results (for, example https://developer.apple.com/forums/thread/126095?answerId=395382022#395382022). 

Basically, with WkWebView, you have to give it read access to the directory the resources are in, which you can do by calling `webview.LoadFileUrl`. 

The first parameter is a uri that must exist, and the second is the uri that file/directory at the first uri should have access too. I've found that calling it with both parameters set to the same thing works. After calling something like `webview.LoadFileUrl(baseURI, baseURI)`, subsequent calls to `webview.LoadHtmlString(html, baseURI)` should work.

Changing the the the Mac handler's LoadHtml() from this

https://github.com/picoe/Eto/blob/d51ed9c561800b444403f1fe7f307885a9dc7e93/src/Eto.Mac/Forms/Controls/WKWebViewHandler.cs#L291-L294

to something like the following should work:

```csharp
                public void LoadHtml(string html, Uri baseUri)
		{
			Control.LoadFileUrl(baseUri.ToNS(), baseUri.ToNS());
                        Control.LoadHtmlString(html, baseUri.ToNS());
		}
```

I didn't have the time to build Eto to test it, and I couldn't override LoadHtml() since it wasn't virtual, but calling `Control.LoadFileUrl(baseUri.ToNS(), baseUri.ToNS());` in a custom Mac handler's overridden `Initialize()` where baseUri was the same in later calls to LoadHtml() did work. (I hope that makes sense).

Another thing to note is that the baseURI on Mac must have a trailing `/`, otherwise the last element seems to get cut off .

I've tested it with setting the baseUri to `Eto.EtoEnvironment.GetFolderPath(Eto.EtoSpecialFolder.ApplicationResources)` with "/wwwroot/" appended. It then works on Mac, and I tested to make sure it didn't break the Wpf target as well.

Miepee:
This just used the info from #2380, I don't currently have access to a Mac, so I didn't test this.

@somelinguist can you (or someone else) use the action-built test suite from here to see whether the fix correctly works?

Fixes #2380  (hopefully)

### DIFF
--- a/src/Eto.Mac/Forms/Controls/WKWebViewHandler.cs
+++ b/src/Eto.Mac/Forms/Controls/WKWebViewHandler.cs
@@ -290,7 +290,10 @@ namespace Eto.Mac.Forms.Controls
 
 		public void LoadHtml(string html, Uri baseUri)
 		{
-			Control.LoadHtmlString(html, baseUri.ToNS());
+			var baseNSUrl = baseUri.ToNS();
+			if (baseNSUrl != null)
+				Control.LoadFileUrl(baseNSUrl, baseNSUrl);
+			Control.LoadHtmlString(html, baseNSUrl);
 		}
 
 		public void Stop() => Control.StopLoading();


cwensley:
Hey @Miepee, thanks for the fix!  This is great, however I think you are missing a null check on `baseUri` before using it, which is its default value.

Miepee:
As far as I can see, the `ToNS()` extension method already includes the null check.
```cs
		public static NSUrl ToNS(this Uri uri)
		{
			return uri == null ? null : new NSUrl(uri.AbsoluteUri);
		}
```

cwensley:
While that is true, the `WKWebView.LoadFileUrl()` call does not allow null for either of its parameters, so it will crash.

PR #2458, opened 2023-04-03
Make use of the virtual method `CreateControl` on wpf `DrawableHandler`

eng-myousif:
* Provide a generic version of `DrawableHandler` to be able to provide custom control, widget and callback.

* Make use of the virtual method `CretaeControl` when creating the wpf canvas for the drawable control.

* This is required to enable the extending the drawable control when needed by a custom canvas.

* That was required in our case because we needed to provide a custom automation peer for the wpf canvas. and we could only do that by providing a custom canvas to the drawable.

* To make use the `EtoMainCanvas` I had to make it public.

### DIFF
--- a/src/Eto.Wpf/Forms/Controls/DrawableHandler.cs
+++ b/src/Eto.Wpf/Forms/Controls/DrawableHandler.cs
@@ -186,10 +186,7 @@ namespace Eto.Wpf.Forms.Controls
 			Control.Loaded += Control_Loaded;
 		}
 
-		public void Create()
-		{
-			CreateControl();
-		}
+		public void Create() { }
 
 		public void Create(bool largeCanvas)
 		{
--- a/src/Eto.Wpf/Forms/Controls/DrawableHandler.cs
+++ b/src/Eto.Wpf/Forms/Controls/DrawableHandler.cs
@@ -11,7 +11,12 @@ using Eto.Wpf.Drawing;
 
 namespace Eto.Wpf.Forms.Controls
 {
-	public class DrawableHandler : WpfPanel<swc.Canvas, Drawable, Drawable.ICallback>, Drawable.IHandler
+	public class DrawableHandler : DrawableHandler<swc.Canvas, Drawable, Drawable.ICallback> { }
+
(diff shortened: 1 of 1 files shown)


eng-myousif:
Hi @cwensley, 
For the other pull request I've noticed there go through checks like `build_windows` and `build_mac` - the third tab in the pull request -  but it is not the case in my pul request. Do I need to do something from my side in that regards?  

Also I would like to highlight that this pull request should help us to be able to automate some important part in our product. Hence it is so appreciated to have a look into my pull request,  thanks in advance!

src/Eto.Wpf/Forms/Controls/DrawableHandler.cs
```diff
@@ -163,18 +168,29 @@ public override void OnUnLoad(EventArgs e)
...
+		protected override void Initialize()
+		{
+			base.Initialize();
+
 			Control.Loaded += Control_Loaded;
 		}
 
+		public void Create()
+		{
+			CreateControl();
```
cwensley: This line should be removed, CreateControl is called automatically.

cwensley (changes requested):
Hey @eng-myousif, thanks for taking the time to put this together!  I think after the one change this should be good to go.

src/Eto.Wpf/Forms/Controls/DrawableHandler.cs
eng-myousif: Thanks for the comment, Applied.

cwensley: Great, thank you!

cwensley (approved):

eng-myousif:
Hi @cwensley , Thanks for merging the suggested enhancement.

I see the assigned milestone is `2.8.0` and I am expecting this will take some time to be released. 
Is it possible to have this enhancement in a patch release then we get it sooner as we needed asap?

cwensley:
@eng-myousif absolutely, that is not a problem.  I just pushed out 2.7.5.  Enjoy!

Issue #2361, opened 2022-12-04
GTK crashes with libwebkit2gtk-5.0


Miepee:
## Expected Behavior
The program should not crash

## Actual Behavior
The program crashes
```
*****Stack Trace*****

   at Eto.GtkSharp.NativeMethods.NMLinux.webkit_web_view_new()
   at Eto.GtkSharp.NativeMethods.webkit_web_view_new()
   at Eto.GtkSharp.Forms.Controls.WebViewHandler..ctor()
   at Eto.GtkSharp.Platform.<>c.<AddTo>b__12_61()
   at Eto.Widget..ctor()
   at Eto.Forms.BindableWidget..ctor()
   at Eto.Forms.Control..ctor()
   at Eto.Forms.WebView..ctor()
```

## Steps to Reproduce the Problem

  1.  Download libwebkit2gtk-5.0
  2. Run a GTK Eto program which uses webkit
  
  

## Specifications

  - Version: 2.7.1
  - Platform(s): GTK
  - Operating System(s): Arch with Flatpak, SteamOS with Flatpak, Gentoo

Miepee:
FWIW, i tried modifying eto to load webkit 5.0 and was greeted with this
``` 
(Eto.Test.Gtk:270719): GLib-GObject-CRITICAL **: 23:53:43.637: specified class size for type 'WebKitWebViewBase' is smaller than the parent type's 'GtkWidget' class size

(Eto.Test.Gtk:270719): GLib-CRITICAL **: 23:53:43.637: g_once_init_leave: assertion 'result != 0' failed

(Eto.Test.Gtk:270719): GLib-GObject-CRITICAL **: 23:53:43.637: g_type_register_static: assertion 'parent_type > 0' failed

(Eto.Test.Gtk:270719): GLib-CRITICAL **: 23:53:43.637: g_once_init_leave: assertion 'result != 0' failed

(Eto.Test.Gtk:270719): GLib-GObject-CRITICAL **: 23:53:43.637: g_object_new_valist: assertion 'G_TYPE_IS_OBJECT (object_type)' failed

(Eto.Test.Gtk:270719): Gtk-CRITICAL **: 23:53:43.637: gtk_container_add: assertion 'GTK_IS_WIDGET (widget)' failed

(Eto.Test.Gtk:270719): GLib-GObject-CRITICAL **: 23:53:43.638: invalid (NULL) pointer instance

[...]

PR #2482, opened 2023-05-19
Fixes Eto.Gtk webview crashing with webkit2gtk-4.1

Miepee:
This PR
- Fixes webviews crashing on Eto.Gtk if the system only has webkit2gtk-4.1 installed instead of webkit2gtk-4.0. (4.1 is API compatible with 4.0, unlike 5.0 which currently hinders #2361)
- Allows net6.0 as an alternative TFM for Eto.Gtk and Eto.Test and fixes the System.Range vs Eto.Forms.Range conflicts that arose from it.


cwensley:
Awesome, thanks for the PR!

PR #2486, opened 2023-05-20
Use Ayatana as an alternative to libAppIndicator

Miepee:
Allows the [Ayatana fork of libappindicator](https://github.com/AyatanaIndicators/libayatana-appindicator) to load as well for system trays.
Obseletes #2055


cwensley:
Looks great!  Thanks for the fix!

PR #2510, opened 2023-07-01
Gtk: Fix drag and drop

### LINKED ISSUE #2492: Gtk drag-and-drop from external applications does not work
## Expected Behavior

The `Control.OnDragEnter`, `Control.OnDragDrop` and friends should allow drag data introspection and retrieval. E.g. if `DragEventArgs.Data.ContainsUris` is `true` then `DragEventArgs.Data.Uris` should be a non-empty array of URIs.

## Actual Behavior

An attempt to access the `DragEventArgs.Data.Uris` property crashes with a `NullReferenceException`.

## Steps to Reproduce the Problem

You can use my project https://github.com/snake-scaly/PinBoard to reproduce the issue. It should support dragging and dropping images from a file manager, an image viewer, or a web browser onto the canvas. The most straightforward test is to drag an image from a file manager and drop it onto the application.

See the `BoardView` class for the relevant code. For me it is 100% reproducible, AKA never works.

## Analysis

The problem is that Gtk drag-and-drop model is arguably incompatible with Eto architecture. This is not a design flaw on the Eto part, this is just the way things are. The main problems are:

* In Gtk, drag data is not available immediately in a `drag-motion` or a `drag-drop` event. An application is expected to call `Gtk.DragDrop.GetData()`, exit the current handler, wait for the `drag-data-received` event, and complete the handling there. There is no such concept in Eto however. The data must be available on demand right in the `OnDragEnter` and `OnDragDrop` handlers. The current solution is to block the handler execution whenever data is requested and spin a nested message loop until `drag-data-received` arrives. The problem with that is that any other events will be processed, too, and if another drag-related event arrives it can mess up the handler's internal state. That other event is probably another `drag-motion` event, I'm not entirely sure.
* In Gtk, an application always receives a `drag-leave` event before getting a `drag-drop` event. It may also never receive the `drag-drop` event if `drag-leave` was generated by the mouse leaving the control boundaries. In the `drag-leave` handler there is no way to tell if a `drag-drop` will follow. Currently the Eto `drag-leave` handler performs some cleanup which *may* be the reason why the application crashes. Again, not entirely sure.

## Solution

https://github.com/picoe/Eto/compare/develop...snake-scaly:Eto:feature/drag-drop-fix

I've made a patch that solves this issue for me. My approach is to request and cache data for all targets in the context, in the normal message loop, before calling `OnDragEnter`. Then I destroy that data on `drag-leave`, and request it again if `drag-drop` happens. This is very resource-intensive, because if there are 7 targets in the context, all 7 will be requested and stored even though many are duplicates, and most will never be requested by the application. This however allows me to stay within the normal message loop and therefore have a predictable state machine in the handler.

## Specifications

  - Version: develop 07989404977ee0a414a7fff7c48a1374bf7faef3
  - Platform: Gtk
  - Operating System: Arch Linux/KDE, regularly updated

snake-scaly:
Ensures that all drop data is immediately available to handlers.

Thisi PR resolves #2492.

### DIFF
--- a/src/Eto.Gtk/Forms/DataObjectHandler.cs
+++ b/src/Eto.Gtk/Forms/DataObjectHandler.cs
@@ -1,4 +1,3 @@
-using Eto.GtkSharp.Drawing;
 namespace Eto.GtkSharp.Forms
 {
 	public class DataFormatsHandler : DataFormats.IHandler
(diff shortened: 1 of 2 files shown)


cwensley:
Hey @snake-scaly, thanks for your contribution!  It is really appreciated.

PR #2534, opened 2023-07-26
Wpf: Fix performance setting TextBox.Text rapidly

cwensley:
This works around a problem in WPF, as reported in https://github.com/dotnet/wpf/issues/5887.  Preventing any GC operations while setting the Text property appears to solve the issue.


JensNordenbro:
@cwensley, Nice!

PR #2574, opened 2023-11-07
Add capability for styling DocumentControl tabs

eng-myousif:
This PR is aiming to expose some feature for `DocumentControl` the following:
1- It is possible  to set the Background and Foreground colors for Tabs and Close ( `x` ) button.
2- It is possible to cancel a closing a page when attempt to close it using close ( `x` ) button.

### DIFF
--- a/src/Eto/Forms/Controls/DocumentControl.cs
+++ b/src/Eto/Forms/Controls/DocumentControl.cs
@@ -22,6 +22,27 @@ public class DocumentPageEventArgs : EventArgs
 }
 
 /// <summary>
+/// Arguments for the <see cref="DocumentControl"/> to get the current page.
+/// </summary>
+public class DocumentPageClosingEventArgs : CancelEventArgs
+{
+	/// <summary>
+	/// Gets the document page.
+	/// </summary>
+	/// <value>The document page.</value>
+	public DocumentPage Page { get; private set; }
+
+	/// <summary>
+	/// Initializes a new instance of the <see cref="T:Eto.Forms.DocumentPageClosingEventArgs"/> class.
+	/// </summary>
+	/// <param name="page">Page.</param>
+	public DocumentPageClosingEventArgs(DocumentPage page)
+	{
+		Page = page;
+	}
+}
+
+/// <summary>
 /// Arguments for the <see cref="DocumentControl"/> when reordering pages.
 /// </summary>
 public sealed class DocumentPageReorderEventArgs : DocumentPageEventArgs
@@ -70,6 +91,7 @@ public class DocumentControl : Container
 	public override IEnumerable<Control> Controls => pages ?? Enumerable.Empty<Control>();
 
 	static readonly object PageClosedEvent = new object();
+	static readonly object PageClosingEvent = new object();
 	static readonly object SelectedIndexChangedEvent = new object();
 
 	/// <summary>
@@ -84,6 +106,15 @@ public class DocumentControl : Container
 	{
 		add { Properties.AddEvent(PageClosedEvent, value); }
 		remove { Properties.RemoveEvent(PageClosedEvent, value); }
+	}
+
+	/// <summary>
+	/// Occurs when the <see cref="DocumentPage"/> is closing.
+	/// </summary>
+	public event EventHandler<DocumentPageClosingEventArgs> PageClosing
+	{
+		add { Properties.AddEvent(PageClosingEvent, value); }
+		remove { Properties.RemoveEvent(PageClosingEvent, value); }
 	}
 
 	/// <summary>
@@ -114,6 +145,15 @@ public class DocumentControl : Container
 		var page = e.Page;
 		if (page != null)
 			page.TriggerClose(e);
+	}
+
+	/// <summary>
+	/// Raises the <see cref="PageClosing"/> event.
+	/// </summary>
+	/// <param name="e">Event arguments.</param>
+	protected virtual void OnPageClosing(DocumentPageClosingEventArgs e)
+	{
+		Properties.TriggerEvent(PageClosingEvent, this, e);
 	}
 
 	/// <summary>
@@ -351,6 +391,11 @@ public class DocumentControl : Container
 		void OnPageClosed(DocumentControl widget, DocumentPageEventArgs e);
 
 		/// <summary>
+		/// Raises the page closing event.
+		/// </summary>
+		void OnPageClosing(DocumentControl widget, DocumentPageClosingEventArgs e);
+
+		/// <summary>
 		/// Raises the selected index changed event.
 		/// </summary>
 		void OnSelectedIndexChanged(DocumentControl widget, EventArgs e);
@@ -373,6 +418,15 @@ public class DocumentControl : Container
 		{
 			using (widget.Platform.Context)
 				widget.OnPageClosed(e);
+		}
+
+		/// <summary>
+		/// Raises the page closing event.
+		/// </summary>
+		public void OnPageClosing(DocumentControl widget, DocumentPageClosingEventArgs e)
+		{
+			using (widget.Platform.Context)
+				widget.OnPageClosing(e);
 		}
 
 		/// <summary>
--- a/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
+++ b/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
@@ -8,6 +8,7 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 	List<DocumentPage> pages = new List<DocumentPage>();
 	ThemedDocumentPageHandler tabPrev, tabNext;
 
+	bool allowNavigationButtons;
 	PointF mousePos;
 	int selectedIndex;
 	float nextPrevWidth;
@@ -41,6 +42,21 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 	{
 		get => Widget.Properties.Get<Padding?>(TabPadding_Key) ?? DefaultTabPadding;
 		set => Widget.Properties.Set(TabPadding_Key, value, DefaultTabPadding);
+	}
+
+	/// <summary>
+	/// Gets or sets a value indicating the tabs can be navigated by next and previous buttons.
+	/// </summary>
+	/// <value><c>true</c> if can be navigated by next and previous buttons; otherwise, <c>false</c>.</value>
+	public bool AllowNavigationButtons
+	{
+		get { return allowNavigationButtons; }
+		set
+		{
+			allowNavigationButtons = value;
+			Calculate();
+			tabDrawable.Invalidate();
+		}
 	}
 
 	/// <summary>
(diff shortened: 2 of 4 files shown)


src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
```diff
@@ -47,16 +71,152 @@ public Font Font
...
+	/// Gets or sets the disabled foreground color.
+	/// </summary>
+	/// <value>The disabled foreground color.</value>
+	public Color DisabledForegroundColor
+	{
+		get { return disabledForegroundColor; }
+		set
+		{
+			disabledForegroundColor = value;
+			Calculate();
```
ievgen-baida: We don't need to `Calculate()` after setting a color as it does not affect any of the sizes being calculated.

src/Eto/Forms/Controls/DocumentControl.cs
```diff
@@ -441,6 +575,66 @@ public new interface IHandler : Container.IHandler
 		/// </summary>
 		/// <value><c>true</c> to allow reordering; otherwise, <c>false</c>.</value>
 		bool AllowReordering { get; set; }
+
+		/// <summary>
+		/// Gets or sets a value indicating the tabs can be navigated by next and previous buttons.
+		/// </summary>
+		/// <value><c>true</c> to allow navigating; otherwise, <c>false</c>.</value>
+		bool AllowNavigating { get; set; }
```
ievgen-baida: Even if `AllowNavigating` is set to `false` the control still allows to navigate, which makes the name of this property a bit misleading. How about renaming to something more specific, like `NavigationButtonsVisible`?

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
```diff
@@ -417,8 +584,11 @@ void Drawable_Paint(object sender, PaintEventArgs e)
 
 		posx = 0;
 
-		DrawTab(g, tabPrev, -1);
-		DrawTab(g, tabNext, -1);
+		if (allowNavigating)
```
ievgen-baida: Those `<` and `>` buttons allow to switch to tabs that are not visible due to the width constraint. Would be nice to be able to scroll to those tabs or have some kind of an overflow menu (perhaps in a different PR)

src/Eto/Forms/Controls/DocumentControl.cs
```diff
@@ -3,7 +3,7 @@
 /// <summary>
 /// Arguments for the <see cref="DocumentControl"/> to get the current page.
 /// </summary>
-public class DocumentPageEventArgs : EventArgs
+public class DocumentPageEventArgs : CancelEventArgs
```
ievgen-baida: Not all events using `DocumentPageEventArgs` support canceling, however when we derive all of them from `CancelEventArgs` it might give a false indication that they are. 

How about creating a separate class just for `DocumentPageClosingEventArgs`?

src/Eto/Forms/Controls/DocumentControl.cs
```diff
@@ -198,6 +217,107 @@ public bool AllowReordering
 		set { Handler.AllowReordering = value; }
 	}
 
+
+	/// <summary>
+	/// Gets or sets a value indicating the tabs can be navigated by next and previous buttons.
+	/// </summary>
+	/// <value><c>true</c> to allow navigating; otherwise, <c>false</c>.</value>
+	public bool AllowNavigating
```
ievgen-baida: Would be nice to add the checkbox for playing with this property into Samples app.

src/Eto/Forms/Controls/DocumentControl.cs
eng-myousif: Or maybe `HasNavigationStepper`, What do you think @ievgen-baida ?

src/Eto/Forms/Controls/DocumentControl.cs
eng-myousif: Applied your comment, you can have a look in this [commit](https://github.com/picoe/Eto/pull/2574/commits/771ed9ece1d1b7c38ed7c77e568df036043cb6b0)

src/Eto/Forms/Controls/DocumentControl.cs
eng-myousif: Applied your comment, you can have a look in this [commit](https://github.com/picoe/Eto/pull/2574/commits/1f4f36902896e7def2622c8ef7c0220ad62a57eb)

src/Eto/Forms/Controls/DocumentControl.cs
ievgen-baida: `HasNavigationButtons` is simpler.

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
eng-myousif: Applied.

src/Eto/Forms/Controls/DocumentControl.cs
eng-myousif: Renamed

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
ievgen-baida: Looks like you went a little to far and also removed this call in `HasNavigationButtons` setter. That one actually influences calculations.

src/Eto/Forms/Controls/DocumentControl.cs
```diff
@@ -198,6 +238,107 @@ public bool AllowReordering
...
+	/// Gets or sets a value indicating the tabs can be navigated by next and previous buttons.
+	/// </summary>
+	/// <value><c>true</c> to allow navigating; otherwise, <c>false</c>.</value>
+	public bool HasNavigationButtons
+	{
+		get { return Handler.HasNavigationButtons; }
+		set { Handler.HasNavigationButtons = value; }
+	}
+
+		/// <summary>
```
ievgen-baida: Minor formatting issue.

test/Eto.Test/Sections/Controls/DocumentControlSection.cs
```diff
@@ -77,6 +77,13 @@ Control SelectPage()
 			return control;
 		}
 
+		Control ShowNavigationStepper()
+		{
+			var control = new CheckBox { Text = "Show navigation buttons" };
+			control.Bind(box => box.Checked, tabControl, documentControl => documentControl.HasNavigationButtons);
```
ievgen-baida: Would be nice to align the naming in this method based on `HasNavigationButtons` property.

ievgen-baida (changes requested):

src/Eto/Forms/Controls/DocumentControl.cs
eng-myousif: Applied

test/Eto.Test/Sections/Controls/DocumentControlSection.cs
eng-myousif: Renamed

cwensley:
Hey @eng-myousif, thanks a ton for the PR!  This looks great.  

A few things, first I don't think the color properties should be on the `DocumentControl` itself, only on `ThemedDocumentControlHandler`, which one could use a style or set directly with that class.  I don't foresee these ever being able to be supported in the same way with any other implementation.

Additionally, it would be great to see the `PageClosing` event and `HasNavigationButtons` implemented for the Gtk handler if possible.

As for the name of `HasNavigationButtons` I would consider using an `Allow` or `Use` prefix, as it's something you can set instead of just query.

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
```diff
@@ -573,8 +575,11 @@ void DrawTab(Graphics g, ThemedDocumentPageHandler tab, int i)
 		g.FillRectangle(backcolor, tabRect);
 		if (tab.Image != null)
 		{
-			var imageSize = tab.Image.Size;
-			g.DrawImage(tab.Image, tabRect.X + TabPadding.Left + (maxImageSize.Width - imageSize.Width) / 2, (tabDrawable.Height - imageSize.Height) / 2);
+			g.SaveTransform();
+			g.ImageInterpolation = ImageInterpolation.High;
+			var imageSize = tab.ImageSize.IsEmpty ? tab.Image.Size : tab.ImageSize;
```
ievgen-baida: Having `ImageSize` public property and setting it to a deliberately "wrong" value can end up in less-than-beautiful results. I would argue the desired image size should be lead by the `ThemedDocumentControl` itself (it knows the dimensions of the tabs it renders and it also should know the ideal dimensions for the image inside that tab).

Shall we just scale `tab.Image` up/down to fit that ideal rectangle?

eng-myousif: I got your point, I will revert changes for exposing ImageSize and let you know.

test/Eto.Test/Sections/Controls/DocumentControlSection.cs
eng-myousif: As per naming convention by Curtis in the [following review comment](https://github.com/picoe/Eto/pull/2574#issuecomment-1802369123). I have renamedit to `AllowNavigationButtons`.

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
eng-myousif: Hi @cwensley ,

- I moved the color properties to `ThemedDocumentControlHandler`

- Implemented `PageClosing` for the Gtk handler.

- Couldn't implement `HasNavigationButtons` (now called `HasNavigationButtons`) for the Gtk handler as the gtk Nootbook control doesn't allow to hide the `HasForwardStepper` and `HasBackwordStepper` because they are readonly, Hence I move the property to the `ThemedDocumentControlHandler`

eng-myousif: Hi @ievgen-baida ,
I have introduce a property `UseFixedTabHeight` along with a checkbox in the DocumentControlSection demo dialog. When set `UseFixedTabHeight` to true it resizes the images inside the Tab height, then we have a fixed tab height that is calculated by the font size and vertical padding.

ievgen-baida: Looks good to me.

Only thing, I wonder if instead of relying on the font size when calculating the height of the image we would rather use some sensible standard size, like 16x16?

eng-myousif: @ievgen-baida, sounds like a good idea. Applied.

ievgen-baida: Thank you. Looks good to me.

cwensley:
Hey @eng-myousif, this looks great!  Let me know when it's ready and I'll merge in the PR.

ievgen-baida:
Hi @cwensley, we finalized it from our side. You can merge it.

Would it be possible to create a patch release, so that we could use it in production?

cwensley:
Hey @ievgen-baida, yes I can certainly create a patch release.  I'll try to do that shortly.

PR #2593, opened 2023-12-08
Prevent SubMenuItem action on macOS

NikolaMendix:
Click on SubMenuItem will close Context Menu or Main Menu. It is not expected behaviour for macOS.

Eto submenuitems

### DIFF
--- a/src/Eto.Mac/Forms/Menu/ButtonMenuItemHandler.cs
+++ b/src/Eto.Mac/Forms/Menu/ButtonMenuItemHandler.cs
@@ -34,7 +34,8 @@ namespace Eto.Mac.Forms.Menu
 		{
 			Enabled = true;
 			Control.Target = new MenuActionHandler{ Handler = this };
-			Control.Action = MenuActionHandler.selActivate;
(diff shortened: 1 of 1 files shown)


cwensley:
Hey this fix looks great, but would you be able to get it set up so that when you subscribe to the Click event it would wire it up again?

NikolaMendix:
Unfortunately no, but if we really need some action on submenu we can use `ButtonMenuItem` instead of `SubMenuItem`? 

Submenu and buttonmenu

NikolaMendix:
Hey @cwensley, any plans for this?

NikolaMendix:
@cwensley, sorry for spamming. Are you ok with this change?

cwensley:
Sorry, `SubMenuItem` will need to support the `Click` event as well otherwise it will break existing code.  This needs to be written in a way to hook up the action when the event is wired up (in `AttachEvent`) so it can work both ways.  The existing `Click` event needs to be translated into a handler event the same way as `SubMenuItem.Opening` has been done so this can be wired up.

PR #2598, opened 2023-12-11
WinForms: Handle MainForm changes

cyanfish:
This allows the former MainForm to be closed without the application shutting down.

Running with a swf.ApplicationContext with a form is equivalent to just running with the form ([reference source](https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/Application.cs,1485)), and allows that form to be changed later.

I'm pretty sure the MainForm.Loaded check is unnecessary as the form should always be loaded when it gets to this point of the Eto.Application.Run flow.


cwensley:
Awesome!  Thanks for the contribution.

PR #2599, opened 2023-12-11
WinForms: Show icon when set in Dialog

cyanfish:
In DialogHandler ShowIcon is set to false:
https://github.com/picoe/Eto/blob/5fb78c03d5c55043b95951640d69030e93c94eeb/src/Eto.WinForms/Forms/DialogHandler.cs#L18

If an icon is manually set it should be displayed.

Using an "if" instead of just setting to true to avoid unnecessary redrawing.


cwensley:
Great, thanks for the contribution!

Would it make sense to set ShowIcon = false if it was then set to null?  (I guess the code would have to be updated to allow a null icon too).

cyanfish:
Yeah, we'd need to update WinForms/WPF/GTK to handle nulls. I could do it if you want, but IMO an Icon is generally a set-once-at-construction thing and I don't think there'd be much use for that.

PR #2605, opened 2024-01-05
Improve DocumentControl styling capabilities

ievgen-baida:
This PR makes it possible to adjust the following styling aspects of `DocumentControl`:

- Set custom `BackgroundColor` of the `DocumentControl`;
- Set custom `TabHoverBackgroundColor` and `TabHoverForegroundColorfor` for the tab in Hover state;
- Set rounded corners of certain `CloseCornerRadius` for `X` Close tab button.

Changes in look/default behavior:

- Removes forced 80% opacity of the tab in Hover state (yet retains previous behavior by default);
- Makes `Highlighted` tab colors winning vs `Hover` colors (as this is how most of the tab controls look);
- Adjusts the size of the close button to balance it with the size of the tab icon (16x16).

### DIFF
--- a/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
+++ b/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
@@ -190,9 +190,9 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 	}
 
 	/// <summary>
-	/// Gets or sets the highlight background color for the highlighted (or active) tab.
+	/// Gets or sets the highlight background color for the highlighted or selected tab.
 	/// </summary>
-	/// <value>The highlight background color for the highlighted (or active) tab.</value>
+	/// <value>The highlight background color for the highlighted or selected tab.</value>
 	public Color TabHighlightBackgroundColor
 	{
 		get { return tabHighlightBackgroundColor; }
@@ -232,9 +232,9 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 	}
 
 	/// <summary>
-	/// Gets or sets the highlight foreground color for the highlighted (or active) tab.
+	/// Gets or sets the highlight foreground color for the highlighted or selected tab.
 	/// </summary>
-	/// <value>The foreground color for the highlighted (or active) tab.</value>
+	/// <value>The foreground color for the highlighted or selected tab.</value>
 	public Color TabHighlightForegroundColor
 	{
 		get { return tabHighlightForegroundColor; }
--- a/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
+++ b/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
@@ -523,7 +523,7 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 					}
 					else
 						SelectedIndex = i;
-						
+
 					break;
 				}
 			}
@@ -691,8 +691,7 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 		var tabRect = tab.Rect;
 		var textRect = tab.TextRect;
 		var closerect = tab.CloseRect;
-		var closemargin =  closerect.Height / 3;
-		var size = tabRect.Size;
+		var closemargin = closerect.Height / 3;
 
 		var textcolor = Enabled ? TabForegroundColor : DisabledForegroundColor;
 		var backcolor = TabBackgroundColor;
@@ -725,7 +724,6 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 			g.DrawLine(closeForeground, closerect.X + closemargin, closerect.Y + closemargin, closerect.X + closerect.Width - 1 - closemargin, closerect.Y + closerect.Height - 1 - closemargin);
 			g.DrawLine(closeForeground, closerect.X + closemargin, closerect.Y + closerect.Height - 1 - closemargin, closerect.X + closerect.Width - 1 - closemargin, closerect.Y + closemargin);
 		}
-
 	}
 
 	/// <summary>
--- a/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
+++ b/src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
@@ -21,6 +21,7 @@ public class ThemedDocumentControlHandler : ThemedContainerHandler<TableLayout, 
 	Panel contentPanel;
 	Font font;
 
+	Color backgroundColor;
 	Color disabledForegroundColor;
 	Color closeBackgroundColor;
 	Color closeHighlightBackgroundColor;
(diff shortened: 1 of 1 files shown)


src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
```diff
@@ -715,8 +715,7 @@ void DrawTab(Graphics g, ThemedDocumentPageHandler tab, int i)
 			textcolor = Enabled ? TabHighlightForegroundColor : DisabledForegroundColor;
 			backcolor = TabHighlightBackgroundColor;
 		}
-
-		if (draggingLocation == null && tabRect.Contains(mousePos) && prevnextsel && !closeSelected && Enabled)
+		else if (draggingLocation == null && tabRect.Contains(mousePos) && prevnextsel && Enabled)
```
eng-myousif, line 734: Did you removed the check for `!closeSelected` intentionally? if yes then why?

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
```diff
@@ -172,9 +173,9 @@ public Color TabBackgroundColor
 	}
 
 	/// <summary>
-	/// Gets or sets the highlight background color for the highlighted  tab.
+	/// Gets or sets the highlight background color for the highlighted (or active) tab.
 	/// </summary>
-	/// <value>The highlight background color for the close highlighted tab.</value>
+	/// <value>The highlight background color for the highlighted (or active) tab.</value>
```
eng-myousif: "(or active)" I think the word `active` is not used across the file but instead the word `selected` is used.
I would rephrase it as following, 
`The background color for the **selected** tab` 

This is how it is document by Microsoft documentation.
image

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
```diff
@@ -215,9 +216,9 @@ public Color TabForegroundColor
 	}
 
 	/// <summary>
-	/// Gets or sets the highlight foreground color for the close button.
+	/// Gets or sets the highlight foreground color for the highlighted (or active) tab.
```
eng-myousif: same as this [comment](https://github.com/picoe/Eto/pull/2605#discussion_r1452046713)

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
```diff
@@ -677,7 +677,7 @@ void CalculateTab(ThemedDocumentPageHandler tab, int i, ref float posx)
 			size.Width += textoffset;
 		}
 
-		var closesize = tabDrawable.Height / 2;
+		var closesize = UseFixedTabHeight ? minImageSquareSide : tabDrawable.Height / 2;
```
eng-myousif: I am not sure if the whole commit is required. (it will also change the current look (size) of the close button for those users whom don't opt in for `UseFixedTabHeight` ) 
Because as a user, I might want to let the tab hight grow  based on the contained  image but still have the close button rendered as 16 x 16. But now this is not possible. What do you think @ievgen-baida ?

ievgen-baida: That is the whole point, because currently the `X` was tiny in relation to 16x16 close button.

`UseFixedTabHeight` is a new option introduced in 2.8.2 so it should be all right to fix it in 2.8.3.

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
ievgen-baida: Sure, this is intentional since it is the only change in 8463f809a64be497d1ecffec78f4b4df9c973232.

It offers is better visual feedback for the tab, its background color and the color of the `X` button when the mouse is over the `X`. 

Previously the tab itself was rendered using default color once `X` was under the mouse.
Now it looks like this:

DocumentControl-Tab-HoverEffect

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
ievgen-baida: I would keep mentioning both "highlighted and selected" because it is not obvious that both states are affected by this property.

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
ievgen-baida: Updated in [2be82e2](https://github.com/picoe/Eto/pull/2605/commits/2be82e215805ebc1d82d71b7a07b39d655f14996).

src/Eto/Forms/ThemedControls/ThemedDocumentControlHandler.cs
ievgen-baida: Updated in [2be82e2](https://github.com/picoe/Eto/pull/2605/commits/2be82e215805ebc1d82d71b7a07b39d655f14996).

ievgen-baida:
@cwensley, do you have any comments about the changes in this PR?

cwensley:
Hi, sorry for the delayed response!  This looks great, however the close button seems to be inflated when using fixed height..

Before:
Screenshot 2024-02-10 at 1 23 25 AM

After:
Screenshot 2024-02-10 at 1 23 57 AM

Without fixed height it appears correct:
Screenshot 2024-02-10 at 1 24 18 AM


Is this intentional?

ievgen-baida:
> Hi, sorry for the delayed response! This looks great, however the close button seems to be inflated when using fixed height..
> 
> Before: Screenshot 2024-02-10 at 1 23 25 AM
> 
> After: Screenshot 2024-02-10 at 1 23 57 AM
> 
> Without fixed height it appears correct: Screenshot 2024-02-10 at 1 24 18 AM
> 
> Is this intentional?

Hi, 
Indeed, it has been done intentionally in a separate commit (9735016263316d2de3fb370f9e87e16c334792db).
I have added description of what has been done in the description of the PR as well.

In short, we needed to balance the size of the close button and the X with the size of the icon, otherwise it appears too small using pre-existing logic (33%/33%/33%).

cwensley:
> Hi,
Indeed, it has been done intentionally in a separate commit (https://github.com/picoe/Eto/commit/9735016263316d2de3fb370f9e87e16c334792db).
I have added description of what has been done in the description of the PR as well.
>
> In short, we needed to balance the size of the close button and the X with the size of the icon, otherwise it appears too small using pre-existing logic (33%/33%/33%).

I'm not sure having different size based on fixed height vs not is ideal here, it should behave/look the same regardless imo. Also I'm not a huge fan of the size of the X personally with the new logic, but I understand this is entirely subjective.  Maybe we need to have a better default scale or logic based on its size?

ievgen-baida:
> I'm not sure having different size based on fixed height vs not is ideal here, it should behave/look the same regardless imo. Also I'm not a huge fan of the size of the X personally with the new logic, but I understand this is entirely subjective. Maybe we need to have a better default scale or logic based on its size?

That is another way of doing it.
However, it would introduce potentially unwanted visual differences for people currently using `DocumentControl` without specifying newly introduced `UseFixedTabHeight`. It basically switches  `DocumentControl` into the mode where the default size of the icon and the X is used instead of relying on the actual images' sizes. 
The current default logic allows to use any images, but it does not allow to use higher-resolution images and make use of scaling when rendered on high-DPI screens.

What kind of logic do you have in mind?

ievgen-baida:
Hi @cwensley,

I have implemented an unconditional logic that does not use custom values for margin or size of the close button.

Please have a look.

cwensley:
Hey @ievgen-baida, thanks for the updates!  This looks great.  If anything that scale could be added as another property that can be configured, but at this point just having a consistent size is perfect.

ievgen-baida:
Hi @cwensley, what it your plan for releasing 2.8.3 with these improvements?

cwensley:
Hey @ievgen-baida, it's released!

PR #2608, opened 2024-01-14
Remove BinaryFormatter from DragDropLib

### LINKED ISSUE #2607: BinaryFormatter is insecure and strongly cautioned against
Not a bug, so eschewing the template for now. The [`DragDropLib`](https://github.com/picoe/Eto/blob/51e1798b71a93358883f64f540d1b9dafab2045e/src/Eto.Wpf/CustomControls/DragDropLib.cs#L385) uses the `BinaryFormatter` class. The .NET folks [recommend against this strongly](https://learn.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide) and have even begun disabling it in more recent versions of .NET. It looks like we only use it in this one place, so it shouldn't be super hard to move off of, I'd guess.

jonko0493:
Fixes #2607.

It seems like this still works perfectly fine in Eto.Test (on WPF at least). It should be functionally the same as the old version but using a more secure API.

### DIFF
--- a/src/Eto.Wpf/CustomControls/DragDropLib.cs
+++ b/src/Eto.Wpf/CustomControls/DragDropLib.cs
@@ -382,9 +382,10 @@ namespace System.Runtime.InteropServices.ComTypes
 					if (ManagedDataStamp.Equals(guid))
 					{
 						// Stamp matched, so deserialize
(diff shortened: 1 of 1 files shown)


cwensley:
Looks great!  Thanks for the contribution.

PR #2612, opened 2024-01-29
Mac: Fix universal binary build

jnkvca:
My MacOS universal binary builds were corrupted when the C# project name contained some combination of letters. I went digging in and found the culprit. The `TrimStart()` method does not trim the specified parameter from the string start, but actually trims as much chars from the front of the string as possible as long as the given parameter contains all those chars ([reference](https://learn.microsoft.com/en-us/dotnet/api/system.string.trimstart?view=net-8.0#system-string-trimstart(system-char()))). This is used to trim the leading relative path of a file, so as a consequence the path gets corrupted.

For example, if my project name was `MYPROJECT` trimming `bin/Debug/net6.0/osx-arm64/MYPROJECT.app/` from `bin/Debug/net6.0/osx-arm64/MYPROJECT.app/Contents/MacOS/MYPROJECT` would result in a corrupted path of `cOS/MYPROJECT` instead of the expected `Contents/MacOS/MYPROJECT`. Then my universal binary `MYPROJECT.app` would actually contain a subdirectory `cOS` with `MYPROJECT` binary.


cwensley:
Aha! Great find, thanks for digging into that and fixing the issue!

PR #2626, opened 2024-02-29
Unsaved indicator for DocumentControl

ievgen-baida:
This PR adds possibility to mark individual document page as one having unsaved changes (`HasUnsavedChanges`). Such page would then be rendered with a little round indicator after the name of the document. The color can be configured using `UnsavedBackgroundColor` property.

Also, the calculations and drawing for each tab in `ThemedDocumentControlHandler` have been refactored:
- Introduce `ImageRect` and `UsavedRect`;
- Calculate rectangles once in `CalculateTab` (except for
`closeForegroundRect` which is calculated while drawing it);
- Restructure calculations element by element;
- Each of the element has valid location and size (even though it can be
 empty);
- (Hopefully) simplify calculations by using `PointF` and `RectangleF`
methods;
- Simplify drawing of the close button.

### DIFF
--- a/src/Eto.Gtk/Forms/Controls/DocumentPageHandler.cs
+++ b/src/Eto.Gtk/Forms/Controls/DocumentPageHandler.cs
@@ -89,6 +89,8 @@ namespace Eto.GtkSharp.Forms.Controls
 			set { closeButton.Visible = value; }
 		}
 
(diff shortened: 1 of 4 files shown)


cwensley:
Looks great, @ievgen-baida!  Thanks for the contribution.

PR #2641, opened 2024-04-09
Failed to initialize WebView2 error message spelling

alfarok:
I was messing around with leveraging WebView2 in my Eto application and noticed this while attempting to get WebView2 to properly load.  

I was getting this error due to `Eto.Wpf.Forms.Controls.WebView2InitializationException: Failed to initialze WebView2 ---> System.UnauthorizedAccessException: Access is denied.` which was caused by not setting an accesible cache location for `WEBVIEW2_USER_DATA_FOLDER`.


cwensley:
I wonder if we should at least by default set a sensible location that is usually writable..

Issue #2443, opened 2023-03-14
[GTK] Error when trying to select row in GridView


Serg-Norseman:
## Expected Behavior

Highlight the selected row in the GridView when the left (primary) mouse button is clicked.

## Actual Behavior

I'm using a GridView, when I click (left button, main) on the last row, the selected row is not highlighted. When I right click on the same row, a selection happens.

Not highlighting a line when clicking with the left button occurs only if this line is the only one or the last one in the list. From the first to the penultimate everything works correctly.

## Code that Demonstrates the Problem

Verified with Eto.Test.Gtk.exe - the last one row (9999) is not clickable (but Down key is working).

## Specifications

  - Version: 2.7.1-2.7.3
  - Platform(s): Gtk
  - Operating System(s): Linux Mint

Serg-Norseman:
The problem has two parts:

1. Gtk-CRITICAL ... error occurs when ReloadData(SelectedRow) is called in the GridView.OnSelectionChanged handler - this was done to make OnCellFormatting work well - it does not work when the selected row changes.

2. Not highlighting a line when clicking with the left button occurs only if this line is the only one or the last one in the list. From the first to the penultimate everything works correctly.

Serg-Norseman:
Maybe it's the same problem: #2401 and #2335?

Serg-Norseman:
Debugging in Eto.Test.Gtk.exe showed that this behavior occurs with AllowEmptySelection, call chain:
GridHandler.Widget_MouseDown() -> AllowEmptySelection=true -> !Control.GetPathAtPos() -> UnselectAll();

If AllowEmptySelection is disabled, the last line is selected normally.

It's up to you to decide if it's a bug or not. And I need to decide - is it permissible for my application to turn off AllowEmptySelection... On good, in my application it is impossible to turn off AllowEmptySelection - violation of the logic of the program.

Serg-Norseman:
Given that GetPathAtPos is a Gtk.TreeView method, the question is why does it think that the click was not on an existing row?

cwensley:
Looks like this was fixed in #2643.  Closing

Issue #2401, opened 2023-02-15
[GTK] last item in treegridview is not editable


Miepee:
## Expected Behavior
It should be editable like other items

## Actual Behavior
its not editable

## Steps to Reproduce the Problem

  1. Try to edit the last item (i.e. clicking the checkbox depending on the page) from a treegridview. This can be observed i.e. in the treegridview test in Eto.Test, but also in the Kitchen Sink
 
## Specifications

  - Version: 2.7.3
  - Platform(s): GTK
  - Operating System(s): Arch Linux

cwensley:
Looks like this was fixed in #2643.  Closing

Issue #2335, opened 2022-10-15
Gtk3: Bottom item in a TreeGridView is unable to be expanded


Prism019:
## Expected Behavior
Expanding the bottom item in a TreeGridView will expand when the expansion arrow is clicked, also running the relevant event.

## Actual Behavior
The element is not expanded.
The Expanding event of the TreeGridView is not called.

## Steps to Reproduce the Problem

  1. Click the expand button on the bottom element of a TreeGridView

## Code that Demonstrates the Problem

```
	public partial class MainForm : Form
	{
		// counter for uniquely identifying items in the TreeGridView
		int counter = 0;
		public MainForm()
		{
			// basic stuff
			Title = "Eto Forms TreeGridView Bug";
			MinimumSize = new Size(800, 800);
			
			// create the grid view, add a column to show the text
			var gridView = new TreeGridView();
			Content = gridView;
			gridView.Columns.Add(new GridColumn{HeaderText = "Text", DataCell = new TextBoxCell(0)});

			// create a new item collection, add a CollectionChanged callback to reload the node once added, and add an initial item.
			var gridViewItems = new TreeGridItemCollection();
			gridView.DataStore = gridViewItems;
			gridViewItems.CollectionChanged += (sender, e) => {
				if(e.NewItems == null) {
					return;
				}
			  	foreach(var i in e.NewItems) {
			  		System.Console.WriteLine("Reloading node " + (i as TreeGridItem).Values[0]);
					gridView.ReloadItem(i as TreeGridItem);
				}
			};
			var parentNode = new TreeGridItem{Values = new object[]{"test_parent" + counter}};
			parentNode.Children.Add(new TreeGridItem{Values = new object[]{"test_child" + counter}});
			counter++;
			gridViewItems.Add(parentNode);

			// callback to telegraph when expanding items occurs
			gridView.Expanding += (sender, e) => {
				System.Console.WriteLine("Expanding!");
			};

			// callbacks to manipulate the number of items in the TreeGridView
			gridView.CellClick += (sender, e) => {
				System.Console.WriteLine("Adding!");
				var addedParentNode = new TreeGridItem{Values = new object[]{"test_parent" + counter}};
				addedParentNode.Children.Add(new TreeGridItem{Values = new object[]{"test_child" + counter}});
				counter++;
				gridViewItems.Add(addedParentNode);
			};
			gridView.KeyDown += (sender, e) => {
				System.Console.WriteLine("Removing!");
[...]

PR #2643, opened 2024-04-15
Gtk fix various treeview bugs

LeeARichardson:
**Fix a couple of selection related bugs in the Grid**

If you switched to Multi mode, unselected everything, and switched back to Single then you could no longer select anything. I reproduced this in raw GtkSharp but I can't find the cause so this is a workaround. I don't think this would break anything if the underlying issue were fixed.

If you switched to Multi and back to Single you could no longer unselect anything. This was due to AllowMultipleSelection starting in Single but setting Browse.

---

**Fix inability to select bottom row of GridView**

This was caused by the header height not being taken into consideration. I did the fixes for the above issues in such a way as to remove the need for the mouse click handler altogether. That handler didn't account for unselecting with the keyboard anyway.

---

These fixes also address (I believe):
https://github.com/picoe/Eto/issues/2335
https://github.com/picoe/Eto/issues/2401
https://github.com/picoe/Eto/issues/2443

and potentially some others but there are too many for me to check.

There are many inconsistencies between the platforms for what happens to the selection on switching modes or which events fire under which conditions. I've just tried to be sensible for what was already happening between Eto, Eto.Gtk, and GtkSharp and any other issues need to be addressed separately.

### DIFF
--- a/src/Eto.Gtk/Forms/Controls/GridHandler.cs
+++ b/src/Eto.Gtk/Forms/Controls/GridHandler.cs
@@ -1,5 +1,9 @@
 using Eto.GtkSharp.Forms.Cells;
 using Eto.GtkSharp.Forms.Menu;
+using Gtk;
+using Grid = Eto.Forms.Grid;
+using GridLines = Eto.Forms.GridLines;
+
 namespace Eto.GtkSharp.Forms.Controls
 {
 	class GridHandler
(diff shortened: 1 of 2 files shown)


cwensley:
Hey @LeeARichardson, this looks great!  Thanks for the contribution.  Only thing I'd recommend is checking if the path from `GetCursor()` is null before using it, as it looks like that API can return nulls.. e.g. if there's no data.

src/Eto.Gtk/Forms/Controls/GridHandler.cs
```diff
@@ -138,12 +139,18 @@ protected override void Initialize()
...
 			base.Initialize();
 
-			Widget.MouseDown += Widget_MouseDown;
+			Control.Selection.Changed += (sender, args) =>
+			{
+				// Prevent unselecting last selected item.
+				if (!AllowEmptySelection && AllowMultipleSelection && Control.Selection.CountSelectedRows() == 0)
+				{
+					Control.GetCursor(out var cursorRow, out _);
+					Control.Selection.SelectPath(cursorRow);
```
cwensley: Should also check for nulls here.

LeeARichardson: Whoops. Done.

cwensley:
Great!  Thank you for the quick edits. 🙏

PR #2644, opened 2024-04-16
Fix an event handler leak in `Binding.RemovePropertyEvent`

### LINKED ISSUE #2446: Binding.RemovePropertyEvent does not remove the 'EventHandler' subscription
Hi there,

Recently, I've upgraded to the 2.7.3 version, where I also contributed via this attached [PR](https://github.com/picoe/Eto/pull/2362).

Unfortunately, the leak was not resolved, even with my above contribution. I'll try to explain the reason for that.

In order to reproduce, you can just chose a random class and add in the constructor / some initialization method the following lines:

```
Binding.AddPropertyEvent(viewModel, vm => vm.Checked, OnCheckedChanged);
Binding.RemovePropertyEvent(viewModel, OnCheckedChanged);
```

## Expected Behavior

After writing the above lines, we expect the `OnCheckedChanged `event handler to never be invoked, as we don't have a subscription to it, as we've invoked `Binding.RemovePropertyEvent` for it.

## Actual Behavior

The `OnCheckedChanged ` event handler is still invoked, even after calling `Binding.RemovePropertyEvent` for it (the breakpoint is hit).

## Specifications

  - Version: 2.7.3
  - Platform(s): WPF 
  - Operating System(s): Windows 10, Windows 11 
  
## Reason for the issue:

In 2.7.3, your source code for the `Binding.RemovePropertyEvent`  looks as following:
```
public static void RemovePropertyEvent(object obj, EventHandler<EventArgs> eh)
{
	var helper = eh.Target as PropertyNotifyHelper;
	if (helper != null)
	{
		helper.Changed -= eh;
		helper.Unregister(obj);
	}
}
```
Previously, I've contributed by adding the `helper.Changed -= eh;` line, but surprisingly, as I've mentioned, the leak is still there, and the event handler keeps getting invoked, even after calling this `RemovePropertyEvent `method.

After investigating a bit, I've discovered that the helper is null, because `eh.Target ` is not of type `PropertyNotifyHelper `, but of the type of the class, where I've defined the event handler (i.e. `CustomCheckBox.cs`). Thus, this code:

```
helper.Changed -= eh;
helper.Unregister(obj);
```

never happens, and the subscription is kept alive.

I couldn't find any issue about it, so was wondering if you are aware of that.

A possible direction is to understand why `eh.Target` stopped working for WPF (assuming it used to work).

joerih:
Removing the handler in `RemovePropertyEvent` doesn't work, as the `Target` of the handler is not the `PropertyNotifyHelper` instance; it is the object that the handler is bound to.

Instead, find the correct handler to remove by checking all invocations of the `PropertyChanged` event of the bound object, and comparing the internal handler of the invocation with the method's argument.

Also, add unit tests to verify the correct behavior.

fixes #2446


cwensley:
Hey @joerih, thanks for submitting the fix and especially for the unit tests.

joerih:
@cwensley Hi Curtis. Is there any chance this fix can be included in the upcoming release?

cwensley:
Hey @joerih, sorry for the late response.  There are a few problems I noticed with this, namely the existing API is actually not complete enough to do this properly, as calling RemovePropertyEvent for the same method (but different properties) would remove it for all properties, not just the one you specify (even though your implementation only removes the first one).

This may mean an API break to return a value for `AddPropertyEvent`, or possibly by adding a `AddPropertyEventEx` that returns an object that you would then pass to `RemovePropertyEvent` (this was actually the original intent of these methods, as it is done like this internally in some cases in Eto).  

I've done some changes locally for this and it seems to work.  I'll try to push some changes for you to review shortly.

cwensley:
Ok I've added some changes.  Please take a look to see if I've missed anything.  In particular, the `UnsubscribingToSameEventWillUnsubscribeAll` test would not work as expected previously, but now will remove all bindings to that delegate.  I've changed AddPropertyEvent and added a few new helper APIs:
```
~object AddPropertyEvent(object obj, string propertyName, EventHandler<EventArgs> eh)
~object AddPropertyEvent<T, TProperty>(T obj, Expression<Func<T, TProperty>> propertyExpression, EventHandler<EventArgs> eh)
+RemovePropertyEvent<T, TProperty>(T obj, Expression<Func<T, TProperty>> propertyExpression, EventHandler<EventArgs> eh)
+RemovePropertyEvent(object obj, string propertyName, EventHandler<EventArgs> eh)
```

The return value for `AddPropertyEvent` can be passed into `RemovePropertyEvent(object, EventHandler<EventArgs>)` to specifically remove that binding, or you can use one of the other `RemovePropertyEvent` that specifies the property so it only removes the binding for that particular property.

E.g. one would do something like this:
```
var boolObj = Binding.AddPropertyEvent(bindObject, obj => obj.BoolProperty, Handler);
// event is active
Binding.RemovePropertyEvent(boolObj, Handler);
// event is inactive
```

This is also totally valid:
```
Binding.AddPropertyEvent(bindObject, obj => obj.BoolProperty, Handler);
// event is active
Binding.RemovePropertyEvent(bindObject, obj => obj.BoolProperty, Handler);
// event is inactive
```

The previous behaviour that you fixed up would still work:
```
Binding.AddPropertyEvent(bindObject, "MyProperty", Handler);
// event is active
Binding.RemovePropertyEvent(bindObject, Handler);
// event is inactive
```

However it was ambiguous in this scenario:
```
Binding.AddPropertyEvent(bindObject, "MyProperty", Handler);
Binding.AddPropertyEvent(bindObject, "MyOtherProperty", Handler);
// MyProperty and MyOtherProperty event is active
Binding.RemovePropertyEvent(bindObject, Handler);
// MyProperty event is inactive.  Now with the added changes both property change events are inactive.
```

joerih:
@cwensley Hi Curtis. Thanks for your response and for improving the fix.

Indeed, I struggled to get this working without changing the API - something I didn't want to do because I was not sure about the consequences. But this way it is even better, as far as I can tell it should cover all scenarios.

cwensley:
> Indeed, I struggled to get this working without changing the API - something I didn't want to do because I was not sure about the consequences. But this way it is even better, as far as I can tell it should cover all scenarios.

Yeah, I am hesitant to make ABI breaking changes as well.  I am thinking instead to obsolete the `RemovePropertyEvent(object obj, EventHandler<EventArgs> eh)` API as it is ambiguous, and require you to specify the property when removing it.  We could then change the signature of the `AddPropertyEvent` back to its original state as there would be no need to retain the return value.

joerih:
> I am thinking instead to obsolete the RemovePropertyEvent(object obj, EventHandler<EventArgs> eh) API as it is ambiguous, and require you to specify the property when removing it. We could then change the signature of the AddPropertyEvent back to its original state as there would be no need to retain the return value.

Yes, that would also work well I think. At the moment we implemented a workaround to avoid the memory leak in our application, and it would not be a problem to switch to a new API when we remove this workaround in the future.

PR #2650, opened 2024-04-30
Added docs for DragDrop Events and Key Events

clicky:
I'm using Eto more and more so I'm going to try and improve docs for things that confused me

### DIFF
--- a/src/Eto/Forms/Controls/Control.cs
+++ b/src/Eto/Forms/Controls/Control.cs
@@ -131,7 +131,9 @@ public partial class Control : BindableWidget, IMouseInputSource, IKeyboardInput
 	public const string KeyDownEvent = "Control.KeyDown";
 
 	/// <summary>
-	/// Occurs when a key has been pressed and is down
+	/// Occurs when a key has been pressed and is down.
+	/// Focus is required for this event to be raised. Ensure CanFocus is set to true.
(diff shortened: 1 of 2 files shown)


src/Eto/Forms/Controls/Control.cs
```diff
@@ -131,7 +131,9 @@ protected virtual void OnSizeChanged(EventArgs e)
 	public const string KeyDownEvent = "Control.KeyDown";
 
 	/// <summary>
-	/// Occurs when a key has been pressed and is down
+	/// Occurs when a key has been pressed and is down.
+	/// Focus is required for this event to be raised. Ensure CanFocus is set to true.
```
cwensley, line 135: CanFocus is only on a Drawable.  Other controls intrinsically can get focus like TextBox, TextArea, DropDown, CheckBox, etc when they are Enabled

src/Eto/Forms/Controls/Control.cs
```diff
@@ -1213,11 +1218,11 @@ public Window ParentWindow
 	public IEnumerable<string> SupportedPlatformCommands => Handler.SupportedPlatformCommands;
 
 	/// <summary>
-	/// Specifies a command to execute for a platform-specific command
+	/// Specifies a command to execute for a Mac platform-specific command
```
cwensley, line 1221: This would also be used for any other platform that has similar needs, not just Mac.  Even though it is only useful for Mac at this point.

PR #2668, opened 2024-06-29
Use base key for key event in GTK

### LINKED ISSUE #2645: Weird behavior with key events for some keys
## Expected Behavior
When pressing `SHIFT + ;` I'm expecting `e.Key` to be equal to `Semicolon`

## Actual Behavior
When pressing `SHIFT + ;`, `e.Key` is set to `None`, although `e.KeyChar` contains the correct character.

```
renderControl.KeyDown += (sender, e) => {
    Debug.WriteLine("Key: " + e.Key);
    Debug.WriteLine("KeyData: " + e.KeyData);
    Debug.WriteLine("KeyChar: " + e.KeyChar);
    Debug.WriteLine("----------------------------------");
    HandleKeyDown(this, new XKeyEventArgs(KeyToInt(e.Key), KeyToInt(e.Modifiers)));
    e.Handled = true;
};
```
In this context, `renderControl` is a `Canvas` control with the `CanFocus` property set to `true`.

Here's the output from the code above, after pressing `;` first and then `SHIFT + ;`

```
Key: Semicolon
KeyData: Semicolon
KeyChar: ;
----------------------------------
Key: LeftShift
KeyData: LeftShift, Shift
KeyChar: ￿
----------------------------------
Key: None
KeyData: Shift
KeyChar: :
----------------------------------
```

## Specifications

  - Version: 2.8.3
  - Platform(s): Gtk, WPF works fine and I haven't tried Mac64 or macOS
  - Operating System(s): Windows 11
  
## Addendum

Here's a list of some of the keys that don't seem to work with `SHIFT`:

`0-9`, `[`, `]`, `;`, `'`, `,`, `.`, `/`,`-`,`` ` ``

psyGamer:
GTK seems to already apply modifiers to the sent key. For example, pressing `;` will get the `semicolon` key from GTK. But `Shift + ;` will get the `colon` key from GTK, which isn't mapped in Eto.
This this, it'll now properly report the key event as `Keys.Semicolon` in `e.Key` and `Keys.Shift` in `e.Modifiers`.

This should close #2645

### DIFF
--- a/src/Eto.Gtk/GtkConversions.cs
+++ b/src/Eto.Gtk/GtkConversions.cs
@@ -516,7 +516,15 @@ namespace Eto.GtkSharp
 
 		public static KeyEventArgs ToEto(this Gdk.EventKey args)
 		{
-			Keys key = args.Key.ToEto() | args.State.ToEtoKey();
+			Keys key;
+
+			// Try to use the base key without modifiers already being applied.
+			// For example: Shift+Semicolon would then actually report that in the key event, instead of Shift+None (technically 'colon', but that isn't mapped)
+			Gdk.Keymap.Default.GetEntriesForKeycode(args.HardwareKeycode, out _, out uint[] keyvals);
+			if (keyvals.Length != 0)
+				key = ((Gdk.Key)keyvals[0]).ToEto() | args.State.ToEtoKey();
+			else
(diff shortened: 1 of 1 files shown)


cwensley:
Nice fix!  Thanks for the contribution.

Issue #2513, opened 2023-07-04
Add WinUI 3 platform


cwensley:
Windows appears to be moving over to WinUI 3 as the "modern" way to build apps, and touts this as the framework they use to build Windows components.  It would be nice to have this as a more modern option than using WinForms or WPF, even though it is likely missing a lot of functionality that will be needed to make a fully rounded out platform in Eto.

ghost:
i seriously advise against this since winui 3 is very very slow and Microsoft doesn't care about its issues at all, its the sole reason i moved to linux.

nvella:
WinUI would be a 'nice to have', but in my eyes is not mature enough, nor appropriately provides enterprise/commercial application developers with the ulilitarian controls, visual styling, or flexibility of WinForms and WPF; WinUI suits only certain use-cases, and those don't tend to be thick desktop apps which I've found Eto exceptional for. Personally, I wish MS would take a page out of Apple's book and iterate, not overhaul.

alexvinidiktov:
Eto.Forms apps with WPF and Windows.Forms backends look very dated. I would love to have WinUI 3 as an additional backend.

Is any work being done on this issue?

If not how would one start implementing WinUI support?

cwensley:
With WPF you can create a modern theme but it takes a lot of effort.

Creating a WinUI 3 platform is certainly a lot more effort, but in case someone like yourself would want to take that on I've created a bare bones platform that can create a Form with a Label on it here: https://github.com/picoe/Eto/pull/2652

I'll see what I can do to get that merged in soon so I can start accepting pull requests filling out more of its functionality.

alexvinidiktov:
> I'll see what I can do to get that merged in soon so I can start accepting pull requests filling out more of its functionality.
That's awesome! I hope it will get merged soon.

cwensley:
Ok it's merged!  I've also added a bare-bones implementation of TableLayoutHandler to get things started a bit quicker.  

My plan for WinUI is to utilize the MeasureOverride and ArrangeOverride as much as possible instead of doing what is in WPF where it sets minimum sizes, though perhaps this isn't feasible.  Either way, we can start filling out the other controls now and once we get the basics in we can utilize the Eto.Test UI.

ris-work:
[...]

Issue #1936, opened 2021-04-23
Please add support for Web Assembly and Tizen


Barshan-Mandal:
Please create a shared project for all platforms

cwensley:
Thanks for the request!  Once upon a time there was an ASP.NET Web Forms platform, however it certainly had limitations.  Web Assembly would not have the same problems as it'd run all in the browser and can respond to events more like desktop apps.

I don't have resources to add these platforms but I could certainly help someone keen to get it going.

ris-work:
I think the WinUI3 backend might support WASM and HTML/Web platforms via platform.uno as Uno Platform is an implementation of WinUI. I am not sure to what extent are WinUI components implemented but simple/trivial WinUI samples work as-is on Uno Platform. I am not sure about Tizen, but in case WinUI3 is completed, Uno also supports rendering directly to the framebuffer (for kiosks with limited resources, etc.), might be something that might me easier to port to Tizen, @Barshan-Mandal .

Barshan-Mandal:
understood brother

ris-work:
Related: [@roozbehid/WasmWinforms](https://github.com/roozbehid/WasmWinforms/issues/16)

ris-work:
Okay, now I got something [closer](https://github.com/ris-work/Eto-WinForms-Native-Unix/): I ported Eto.Platform.Windows to use Core.System.Windows.Forms in place of System.Windows.Forms and trivial applications do work (the platform tests start but _switching panels won't work at all_) with native Cairo/libgdiplus backend on modern .net (.net 10). It is in the branch "Core.System.Windows.Forms". I have stubbed all the DllImport/FFI calls with default returns.

WasmWinforms apparently depends on Mono and I don't think it is straightforward anymore as the current target I've made is `net10` and mono only does up to `netfx4xx` and `netstandard2.0`. I have no idea where to go from here, but a WebSockets based X11 renderer (more complex but faster) with something like MicroWindows-wasm or a full x86(-64) emulator like [v86](https://copy.sh/v86/) or [tinyEMU](https://bellard.org/tinyemu/) might be the easiest way to go from here. Or, possible to use a VNC backend with a VNC renderer, still better than x86(-64) emulation but likely will cost server resources.

[...]

Issue #410, opened 2015-11-20
[Req] Native (pure C#) wayland backend support


ArsenShnurkov:
Wayland is a protocol. Protocol can be implemented.

cwensley:
Thanks for the suggestion!  Though, why would we bind directly to wayland when we already use GTK (which sit atop wayland)?  Doesn't this reinvent the wheel?  From what I understand, we'd have to implement all widgets, windowing functionality, etc.

What do you see as the benefit to this approach?

drakon660:
It is good idea, but for mono framework or .net core not for eto framework. The advantages are obvious. Eto -> gtk# -> mono/.net core -> gtk+ -> xwindow. Only eto -> mono/.net core -> wayland.

cwensley:
I could be wrong, but it appears that Wayland just provides a compositer.  It does not appear to provide a drawing api to draw lines, gradients, textures, etc.  This is what something like cairo would provide and does it very well. I don't envision Eto.Forms implementing its own drawing api that renders to a buffer, since this would re-invent the wheel and probably do it poorly.

However, I have often pondered creating custom drawn handlers for each of Eto's controls (such as Button, TextBox, etc) using the Eto.Drawing api.  This would mean that one would only have to wrap the Eto.Drawing handlers for a platform in order to support it (if you don't care about how it looks visually).  

Eto.Forms already wraps cairo (which supports wayland), so it would just be a matter of creating custom-drawn handlers for each control and we can at least cut out GTK+.  This, however, is still a lot of work.  If someone is keen enough, this would be a welcome addition and I could help get this started, but won't be able to do much of the work.

CsharpOnLinuxDev:
Is there anyone still around interested in this idea? I've managed to create a binding to Wayland that enables C# to use Wayland to create User Interface with ease. I plan on finalizing the design before uploading a repo for it.

cra0zy:
I don't see why this issue is even a thing... What would you gain?

CsharpOnLinuxDev:
Eliminating the needs for GTKSharp and GTK dependencies is a big one. Wayland is relatively fast and flexible enough that we could use libraries like Skia, OpenGL, ImageSharp, or other yet to list libraries to render our Eto.Form rather than sticking with very chucky library like GTK, we can go for a smaller and more slim user interface that provides more options to the users especially in embedded environment ( https://www.youtube.com/watch?v=GtXQJ0c5q0k to see how Wayland is used on ARM device).

[...]

PR #2669, opened 2024-07-03
WinUI: More control implementations

cwensley:
Added initial stub controls for some of the basics.  A lot of the functionality is not yet implemented for these so still needs a lot of work.

Cite #2513


ris-work:
Thanks for this! This might be the one that finally makes things available for WASM applications and direct framebuffer rendering through [Platform.Uno](https://platform.uno).

ris-work:
Related: #1936 #410

PR #2671, opened 2024-07-07
Mac: Implement Keymap.Convert

psyGamer:
This should allow you to retrieve the `Shortcut` of `MenuItem`s again.
I can't verify that this actually works since I'm having difficulties with compiling on Mac, but I'm pretty sure it _should_ work.


cwensley:
Looks great!  Thanks again for another great contribution.

PR #2673, opened 2024-07-12
Mac: Don't use colourizer for NSTextField

cwensley:
There's no need to use the colourizer for `NSTextField` as it intrinsically supports background colors, so disable its usage.  We do still need the `DrawInteriorWithFrame` so that macOS draws the background only _within_ the frame instead of behind it.

The colourizer has been known to crash in this case on macOS Sonoma and later.


cwensley:
Actually, closing this as the colourizer is only used when alpha is on, which is still needed as otherwise a transparent background would still show the white background.

PR #2675, opened 2024-08-19
Add WebView.MessageReceived

camnewnham:
Hello! Initial draft for discussion.

The goal of this PR is to create a uniform approach to sending events from JS to Eto. Combined with the existing `ExecuteScript` it should allow for bi-directional communication (without polling).

Since each browser implementation significantly differs this approach is intended to keeps it simple by adding a single object /method to the global JavaScript context.

### Example usage:
```JS
function isRunningInEto() {
  return window.eto !== undefined;
}

function onButtonClick() {
   window.eto.postMessage("The button was clicked!");
}
```

### Supported Implementations:
- [x] WebView (legacy) (Wpf and WinForms)
- [x] WebView2 (Wpf and WinForms)
- [x] WebKit.WKWebView (Mac64 and MacOS)

Not implemented:
- [ ] WebKit.WebView (legacy)
- [ ] GTK

### DIFF
--- a/src/Eto.Mac/Forms/Controls/WKWebViewHandler.cs
+++ b/src/Eto.Mac/Forms/Controls/WKWebViewHandler.cs
@@ -10,7 +10,12 @@ namespace Eto.Mac.Forms.Controls
 	{
 		public override NSView ContainerControl { get { return Control; } }
 
-		public wk.WKWebViewConfiguration Configuration { get => Control.Configuration; }
+		public wk.WKWebViewConfiguration Configuration { get; set; }
+
+		public WKWebViewHandler()
+		{
+			Configuration = CreateConfiguration();
+		}
 
 		protected override wk.WKWebView CreateControl()
 		{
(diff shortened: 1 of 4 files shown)


cwensley:
Hey @camnewnham, this is looking great!  This will help a lot of custom script scenarios.

Do you need help with the GTK implementation or is that something you'll be able to do?

src/Eto.Mac/Forms/Controls/WKWebViewHandler.cs
```diff
@@ -10,7 +10,7 @@ public class WKWebViewHandler : MacView<wk.WKWebView, WebView, WebView.ICallback
 	{
 		public override NSView ContainerControl { get { return Control; } }
 
-		public wk.WKWebViewConfiguration Configuration { get; set; } = new wk.WKWebViewConfiguration();
+		public wk.WKWebViewConfiguration Configuration { get => Control.Configuration; }
```
cwensley: This might break some scenarios where the configuration is set by a derived handler or custom style.  I think we'll still want a setter here for those cases.

camnewnham: Good point. I've reinstated this in https://github.com/picoe/Eto/pull/2675/commits/e4bedc3251db1ce6ff086bdab1043def5aba2497 and used `CreateConfiguration` to create a default value in the constructor.

camnewnham:
> Do you need help with the GTK implementation or is that something you'll be able to do?

That would be appreciated - I don't currently have a Linux machine (or is this testable on other platforms?)

Is a legacy `WebKit.WebView` implementation required?

cwensley:
Hey @camnewnham, thanks for doing the initial work on this!  It's really appreciated.

I've finished the implementation for GTK and added a unit test, so things should now be working on all (major) platforms.

PR #2722, opened 2025-01-15
Allow disposing multiple times

SlowLogicBoy:
I use dependency injection.

As per Disposable pattern `Dispose` method can be called multiple times however, once DependencyInjection calls `Wiget.Dispose` it crashes, because Handler property checks if it's already disposed.

As I understand Application disposes all of it's widgets and cannot be called 2nd time which is done by DependencyInjection.


cwensley:
Looks great, thanks for catching that @SlowLogicBoy!

PR #2797, opened 2025-07-17
Fixes crash when switching open file dialog type filter to *.*

eirannejad:


eirannejad:
Hmm. PR made on the wrong repo I think

PR #2816, opened 2025-09-07
Fix(macOS): Prevent nested app bundle on publish

AkiSakurai:
When publishing a .NET Core application on macOS, the `MacCopyPublishFiles` target would incorrectly copy the generated .app bundle into itself. This resulted in a nested directory structure (e.g., `MyCoolApp.app/Contents/MacOS/MyCoolApp.app`).

This was caused by excluding `$(LauncherPath)` which only excludes the executable, not the entire app bundle directory.

The fix changes the exclusion to `$(OutputAppPath)`, which correctly points to the root of the `.app` bundle, preventing it from being copied into its own `Contents/MacOS` directory.


cwensley:
Ooh!  Nice catch there, thanks for the fix!

PR #2822, opened 2025-09-14
Wpf/WinForms: add text alignment to NumericStepper control

luke-ransfield:
Adds text alignment to the NumericStepper control for WPF/WinForms. Also added examples of NumericStepper for each alignment to the Eto.Test App:

Screenshot 2025-09-14 171843

Have set the default alignment to `TextAlignment.Left` to keep it consistent with existing behavior.

Have made changes for Gtk but haven't tested.


cwensley:
Hey @luke-ransfield,

Thanks so much for submitting this PR.  I've added the NumericStepper.TextAlignment property in #2930 and tested on all platforms to ensure consistent/correct behaviour.

Cheers!
Curtis.

PR #2825, opened 2025-09-17
[WinForms] CanFocus was apparently never implemented, but CanFocusMe is there. D…

ris-work:
…uplicate CanFocusMe because CanFocusMe is used elsewhere. On WinForms, to have an Eto.Drawable receive focus, instead I had to cast it to the native control and set CanFocus. This hopefully fixes this.

This is my first PR in Eto.Forms. Hopefully I got the architecture correctly.

### DIFF
--- a/src/Eto.WinForms/Forms/Controls/DrawableHandler.cs
+++ b/src/Eto.WinForms/Forms/Controls/DrawableHandler.cs
@@ -36,6 +36,12 @@ namespace Eto.WinForms.Forms.Controls
 				set { canFocus = value; SetStyle(swf.ControlStyles.Selectable, value); }
 			}
 
+			public bool CanFocus
+			{
+				get { return canFocus; }
+				set { canFocus = value; SetStyle(swf.ControlStyles.Selectable, value); }
+			}
+
 			protected override void OnGotFocus(EventArgs e)
 			{
 				base.OnGotFocus(e);


cwensley:
Hey @ris-work, thanks for the contribution!  It's really appreciated.  One thing you may have missed is that the `DrawableHandler.EtoDrawable.CanFocusMe` property is set from the [`DrawableHandler.CanFocus`](https://github.com/picoe/Eto/blob/6e521f375a262dd5228a910b88c402bf607fca21/src/Eto.WinForms/Forms/Controls/DrawableHandler.cs#L109) property, which in turn can be set simply by setting `myDrawable.CanFocus = true`.  Adding this property to `EtoDrawable` isn't actually necessary.  Is there something that isn't working for you when setting the `Drawable.CanFocus` property?

ris-work:
Thanks, I think the issue is elsewhere, maybe how controls interact with each other. On the latest one, I could not reproduce the issue with either Winforms or WPF (both work, but the caveat is that WPF's one participates by default with Tab navigation while Winforms needs something else too, to participate in Tab navigation. It is focusable, but by mouse, it receives key events once click-focused; still could not get it to participate in Tab navigation. WPF works alright.).

I am closing this issue as it is an entirely separate issue why it does not participate in Tab-navigation. If anyone has any idea, please let me know.

PR #2845, opened 2025-10-18
Fix typo in RuntimeIdentifier property references in Mac.targets

AkiSakurai:


cwensley:
🤦‍♂️good catch, thanks again!

PR #2861, opened 2025-10-29
Make groupbox padding with empty header symmetric on wpf backend

AkiSakurai:
The padding in WPF GroupBox controls with empty header was asymmetric due to extra height from the Label header. To achieve symmetry, the header is set to null, and a top padding is added, value extracted from the WPF source code.

https://github.com/dotnet/wpf/blob/092d18b5b2a6f0f1d9963730a4423c95c84652cc/src/Microsoft.DotNet.Wpf/src/Themes/XAML/GroupBox.xaml#L33

Original Screenshot:

original

Fixed Screenshoot:
fixed


cwensley:
Sorry this took so long, but this should now be fixed with #3054.

PR #2885, opened 2026-01-06
Centralize Eto.Gtk use of NativeLibrary.SetDllImportResolver

### LINKED ISSUE #2884: Eto.Gtk can erronously call NativeLibrary.SetDllImportResolver twice
Eto.Gtk uses `NativeLibrary.SetDllImportResolver` in [LinuxTrayIndicatorHandler.cs](https://github.com/picoe/Eto/blob/d63474c4dee2f5c19839838168373bda930eacf5/src/Eto.Gtk/Forms/LinuxTrayIndicatorHandler.cs#L54) as well as [NativeMethods.cs](https://github.com/picoe/Eto/blob/d63474c4dee2f5c19839838168373bda930eacf5/src/Eto.Gtk/NativeMethods.cs#L296) in NMLinux, NMWindows and NMMac classes. This function may only be called once per assembly as per the [documentations](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.setdllimportresolver?view=net-10.0). 

> Only one resolver can be registered per assembly. Trying to register a second resolver fails with an [InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-10.0).

I ran into this issue when using `Eto.Forms.TrayIndicator` and `Eto.Forms.Screen.PrimaryScreen.GetImage()`, which led to a TypeInitializationException in the latter (because `NativeLibrary.SetDllImportResolver` threw).

Descolada:
Fixes #2884.

Eto.Gtk previously called `NativeLibrary.SetDllImportResolver` in multiple static constructors (`NMLinux`, `NMWindows`, `NMMac`, `LinuxTrayIndicatorHandler`), but an assembly can only register one resolver—any later registration threw, leaving the type unusable.

This change introduces `DllImportResolverManager` to register a single resolver for the assembly and fan out to any number of per-feature delegates. Access to the delegate list is serialized with a lock for thread safety; resolvers themselves must remain load-only (no Add/Remove during resolution) to avoid mutating the collection while it’s being iterated.

### DIFF
--- a/src/Eto.Gtk/DllImportResolverManager.cs
+++ b/src/Eto.Gtk/DllImportResolverManager.cs
@@ -1,0 +1,52 @@
+#if NET
+namespace Eto.GtkSharp
+{
+	static class DllImportResolverManager
+	{
+		static readonly object gate = new object();
+		static readonly List<DllImportResolver> resolvers = new List<DllImportResolver>();
+
+		static DllImportResolverManager()
+		{
+			NativeLibrary.SetDllImportResolver(typeof(DllImportResolverManager).Assembly, Resolve);
+		}
+
+		public static void Add(DllImportResolver resolver)
+		{
+			if (resolver == null)
+				throw new ArgumentNullException(nameof(resolver));
+
+			lock (gate)
+			{
+				resolvers.Insert(0, resolver); // Insert at beginning so more recent entries get priority over older ones 
+			}
+		}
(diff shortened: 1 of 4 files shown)


cwensley:
Beautiful!  Thanks for the fix! 🙂

PR #2887, opened 2026-01-19
Improve RadioButton Click event implementation

Descolada:
RadioButton is documented to have a [Click event](https://pages.picoe.ca/docs/api/html/E_Eto_Forms_RadioButton_Click.htm), but it apparently wasn't wired up on platforms other than macOS.

Expected behavior: triggers once on every click on the radio button, even if it's already checked.

- [x] WinForms: tested, but for unknown reasons the event triggers twice (I could not determine the reason in a reasonable timeframe and gave up on it)
- [x] Wpf: tested, working
- [x] Gtk: tested, working
- [ ] iOS: not tested
- [ ] macOS: not tested (existing handler unchanged)
- [ ] Android: not tested

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -121,6 +121,7 @@
 			"request": "launch",
 			"preLaunchTask": "build-winforms",
             "program": "${workspaceFolder}/artifacts/test/Eto.Test.WinForms/${config:var.configuration}/net48/Eto.Test.WinForms.exe",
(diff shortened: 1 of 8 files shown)


cwensley:
This looks great!  Thanks for the fixes. 🙏

I've added a fix for WinForms so it doesn't trigger twice, looks like the control styles were causing that, but I've verified with them off it still behaves correctly.

PR #2906, opened 2026-02-26
WinForms: Fix RichTextArea.SetItalic()

acormier:
Calling `RichTextArea.SetItalic()` on WinForms set the bold state, not the italic state.


cwensley:
Thanks @acormier!  I'm surprised that bug wasn't found yet.

PR #2916, opened 2026-03-08
Gtk: avoid duplicate clipboard entries on repeated updates

Descolada:
This fixes a GTK clipboard issue where repeatedly setting clipboard contents could leave the handler with accumulated duplicate target entries and lead to crashes on subsequent reads/updates.

The previous implementation appended new clipboard entries every time Text, Html, Image, Uris, or custom data were set. Over time that caused the GTK target list and the handler’s internal entry list to likely enter a stale state and eventually cause a crash. This change normalizes clipboard state by replacing existing entries of the same logical kind instead of always appending.


cwensley:
Looks great!  Thanks for the fixes. 🙏

PR #2918, opened 2026-03-15
Fix GTK GroupBox to adhere to user-preferred size

Descolada:
**Description**  
Fixes a GTK issue where `GroupBox` could expand beyond its explicit size if a child control had a larger preferred width.

**Change**  
- `GroupBoxHandler`: use `EtoFrame { Handler = this }` instead of plain `Gtk.Frame`
- Add minimal `EtoFrame` wrapper applying `UserPreferredSize` in:
  - `OnGetPreferredWidth`
  - `OnGetPreferredHeight`

**Result**  
Explicit GroupBox size is preserved; oversized children no longer force GroupBox width growth.


cwensley:
Thanks @Descolada!  Looks great.

PR #2921, opened 2026-03-22
Fix broken WithSize on GTK due to image size caching

AkiSakurai:
The WithSize logic was broken on GTK because of the image size cache. During construction of a new icon, the image size could change *after* it had already been accessed and cached, leading to inconsistent or outdated size information.

This commit changes icon construction to fetch the current (non-cached) size directly from the size handler instead of relying on the potentially stale cached value.

The issue can be observed/reproduced using the existing unit test:
  ImageSizeShouldUpdateImageViewSize

https://github.com/picoe/Eto/blob/c426ea7bb32945a4d7db3384e0ed2fef293ae60d/src/Eto.Gtk/Drawing/IconHandler.cs#L179

### DIFF
--- a/src/Eto/Drawing/Icon.cs
+++ b/src/Eto/Drawing/Icon.cs
@@ -205,7 +205,7 @@ public class Icon : Image
 		float lastScale = 0;
 		foreach (var frame in Frames.OrderBy(r => r.Scale).ThenByDescending(r => r.Size.Width * r.Size.Height))
 		{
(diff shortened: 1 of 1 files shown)


cwensley:
Thanks for this, though this might not be the right fix.  I have a PR coming that should fix this in a different way.  Short story is, we shouldn't be using GetFrame() during construction of the icon with frames.

PR #2945, opened 2026-05-15
Mac DropDown: don't force control.Font in DrawTitle on macOS 26+ to a…

acormier:
…void measure/draw mismatch that elides short titles.


cwensley:
Hey @acormier, thanks for the PR, but this is the incorrect fix imo.  This PR will make the drop down show the selected item in the formatted font, which works for (many) fonts.  However, some fonts will overflow the text outside the control because it does not change its size to fit the font (e.g. Zapfino), and in macOS 26/Liquid Glass this is still the case. 

Pre-macOS 26/Liquid Glass, it would show the correct font but the position of the text was outside of the control, which is obviously also incorrect.  The new PR #2948 fixes both issues, and will make the drop down always show the standard font in the control (and still custom fonts in the popup).

PR #2962, opened 2026-06-14
FileDialog: Preserve the current working directory

Descolada:
File selection dialogs should not modify the application's global working directory: this is unexpected and inconsistent between platforms. This PR preserves the current directory across GTK, WinForms, and WPF after a file or folder dialog is closed. Mac, Android, iOS already do not change it, so they were left unchanged.

### DIFF
--- a/src/Eto.Gtk/Forms/GtkFileDialog.cs
+++ b/src/Eto.Gtk/Forms/GtkFileDialog.cs
@@ -108,9 +108,7 @@ namespace Eto.GtkSharp.Forms
 #endif
 
 			DialogResult response = ((Gtk.ResponseType)result).ToEto ();
-			if (response == DialogResult.Ok && !string.IsNullOrEmpty(Control.CurrentFolder))
-				System.IO.Directory.SetCurrentDirectory(Control.CurrentFolder);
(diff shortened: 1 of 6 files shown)


cwensley:
Hey @Descolada, thanks for this.  Did you test the open/save dialogs and whether they remember their current folder if you change it between uses?  E.g. if you change folder to save in a particular location, when you go to save again it was supposed to remember that same location.

Descolada:
@cwensley I did not, but now I have tested: WinForms and WPF use the same file/folder dialog which remembers the last used folder, but Gtk regressed, so I added a commit which fixes that.

cwensley:
That's great, those changes look good.  

I guess one last tiny nitpick.  `Control.RestoreDirectory = true;` for winforms and wpf should be in the `Initialize()` method instead of `ShowDialog()` like so, so it can be changed via a style:

```
		protected override void Initialize()
		{
			base.Initialize();
			Control.RestoreDirectory = true;
		}
```

Descolada:
Done 👍

cwensley:
Beautiful!  Thank you. 🙏

PR #2963, opened 2026-06-14
Mac: Default ShowInTaskbar to true

Descolada:
Mac windows currently default ShowInTaskbar to false, contrary to the documented behavior and other desktop platforms. Default it to true while preserving explicit opt-out.

### DIFF
--- a/src/Eto.Mac/Forms/MacWindow.cs
+++ b/src/Eto.Mac/Forms/MacWindow.cs
@@ -664,11 +664,9 @@ namespace Eto.Mac.Forms
 			}
 		}
 
-		public bool ShowInTaskbar
-		{
-			get;
-			set;
-		}
+		// Defaults to true to match the documented behavior (and the Windows platform handler):
+		// windows show in the taskbar/Dock unless explicitly told not to.
+		public bool ShowInTaskbar { get; set; } = true;
 
 		public bool Closeable
 		{


cwensley:
I don't really think this makes any sense, it doesn't actually show the window in the taskbar as the taskbar on macOS is per-application, not per-window.  So, really it should always return false, regardless whether you set it to true.  Thoughts?

Descolada:
Yeah, unfortunately ShowInTaskbar has different semantics in Mac as it does in other platforms. As far as I know, Eto currently doesn't have a way to toggle the Mac taskbar icon on/off, and as you said it's per-application, not per-window. However, I also believe that `ShowInTaskbar = false` in Mac is also incorrect, because by default Mac *does* show a Dock icon, unless the user has modified .pinfo or set NSApplication.ActivationPolicy to accessory. So, because ShowInTaskbar is by default true in other platforms, and by default Mac also has a dock icon, I thought it would make sense to default it to true. Additionally, it makes it easier to keep track of the *intent* the user has for the window. For example suppose Eto implemented a way to automatically hide the dock icon once all windows are either not visible or ShowInTaskbar is false. In that case by default we might hide the icon (if the user has not explicitly set ShowInTaskbar to true), which would have the opposite effect of what happens by default on other platforms.

cwensley:
Makes sense, thanks for the PR!

PR #2964, opened 2026-06-14
Mac: Bring focused windows to the foreground

Descolada:
MakeKeyAndOrderFront does not activate a background macOS application, so Window.Focus() may fail to foreground the requested window. Activate the application first, then make the window key and bring it forward.

### DIFF
--- a/src/Eto.Mac/Forms/MacWindow.cs
+++ b/src/Eto.Mac/Forms/MacWindow.cs
@@ -944,7 +944,14 @@ namespace Eto.Mac.Forms
 			}
 		}
 
-		public override void Focus() => Control.MakeKeyAndOrderFront(Control);
+		public override void Focus()
+		{
+			// Activate the NSApplication first so the window actually comes to the foreground.
+			// MakeKeyAndOrderFront alone is not enough when the process is in the background
+			// (e.g. after transitioning from Accessory to Regular activation policy).
+			NSApplication.SharedApplication.Activate();
+			Control.MakeKeyAndOrderFront(Control);
+		}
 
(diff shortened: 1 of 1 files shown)


cwensley:
Hey @Descolada, so I've thought about this one, and I think I'd like to see an `Application.Instance.Activate()` API instead of making macOS try to activate every time you focus a window.  Thoughts?

Descolada:
@cwensley interesting idea. As far as I understand, currently only macOS distinguishes active app (the overarching currently used program, claims the global menu bar at the top of the screen, owns the Dock icon indicator etc) vs focused window (where keyboard focus is). It's possible to activate an app without focusing a window, but not focusing a window without activating the app (and a window can't have keyboard focus without also being the active app). So, I think `Application.Instance.Activate()` would have a place to activate the app, but that would likely be a macOS-only feature (I'm not sure how that would work on other platforms - on Windows/Linux what would it activate?). Also, if `Window.Focus()` didn't automatically call it then the semantics of `Window.Focus()` would remain incorrect, because it sometimes won't bring the window to focus due to app-level activation status being wrong.

On further thought, it could perhaps be a property instead `Application.Instance.Active`:
* on macOS the getter would return application-level focus status (supposing it's possible to query it, I haven't looked into it), the setter would set the app-level focus
* on Windows/Linux it could answer the question whether the currently focused window belongs to our `Application.Instance`, but I don't think the setter would have an use

What do you think? If you think this is worth implementing then I can attempt it + change the current PR to do `if (!Application.Instance.Active) Application.Instance.Active = true` before attempting window focusing.

PR #2965, opened 2026-06-14
Add clipboard changed event

Descolada:
Adds a `Clipboard.Changed` event to notify applications when the system clipboard contents change.

Implementations are included for WinForms, WPF, GTK, macOS, and iOS. Manually tested on WPF, WinForms, GTK, and macOS.

Also:
- Adds clipboard change monitoring to the Clipboard test section.
    - **WinForms/WPF:** Registers a hidden native window with `AddClipboardFormatListener` and handles `WM_CLIPBOARDUPDATE`.
    - **GTK:** Uses the clipboard owner-change notification.
    - **macOS:** Polls the pasteboard change count while the event has subscribers.
    - **iOS:** Observes pasteboard change notifications.
- Improves handling of unsupported Windows clipboard formats.
- Fixes Windows DIB stream positioning and premultiplied-alpha conversion.
- Prevents empty WinForms layouts from causing an indexing exception.

### DIFF
--- a/src/Eto.Gtk/Forms/ClipboardHandler.cs
+++ b/src/Eto.Gtk/Forms/ClipboardHandler.cs
@@ -1,7 +1,7 @@
 using Eto.GtkSharp.Drawing;
 namespace Eto.GtkSharp.Forms
 {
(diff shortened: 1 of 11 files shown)


cwensley:
Thanks for the PR! I debated whether this API should be an instance event or static, but I think this is the right way to do it.

PR #2966, opened 2026-06-14
Gtk: Add Wayland clipboard support

Descolada:
> [!IMPORTANT]
> Depends on the clipboard change notifications PR #2965

Adds a native Wayland clipboard backend for GTK.

GTK's standard Wayland clipboard access is focus-gated, which prevents reliable clipboard access and change notifications while the application is unfocused. This implementation uses the `zwlr_data_control_manager_v1` or `ext_data_control_manager_v1` protocol through a dedicated `libwayland-client` connection.

The backend supports:

- Reading and writing text, images, URIs, and custom MIME data.
- Enumerating available MIME types.
- Clearing the clipboard.
- Receiving clipboard change notifications while unfocused.
- Falling back to the existing GTK clipboard implementation when Wayland or the data-control protocol is unavailable.

Manually tested with GTK on Wayland.


cwensley:
Hey @Descolada, thanks for this!  One change I'd like to see is instead of modifying the current ClipboardHandler to support both Wayland and Gtk, could you instead refactor this to have a WaylandClipboardHandler that is chosen during registration instead?  This would make it a lot cleaner and easier to support.

Descolada:
Sorry, took me a while to test the changes. In the process I discovered two bugs:
1. GTK erronously emits multiple owner-change events per clipboard change in Wayland, which is a known unresolved bug (GNOME #775631). I worked around it by debouncing (alternatives include just accepting multiple event triggers per one clipboard change, or fingerprinting clipboard contents and only calling events for actual clipboard changes)
2. GtkSharp has a bug involving EventOwnerChange.Owner struct types. I opened [an issue](https://github.com/GtkSharp/GtkSharp/issues/468) for it, hopefully it'll be fixed.

cwensley:
Nice!!  Thanks yet again @Descolada. 🙏

PR #2977, opened 2026-07-02
Gtk: encode bitmaps in memory instead of a temp file

Descolada:
Not sure why GTK `BitmapHandler.Save` used a path which saved a file to a disk, then read it and deleted, but directly writing to the stream seems to work. Am I missing something here?


cwensley:
No clue.. maybe GTK# 2 had issues back in the day, or didn't have the SaveToBuffer API?  Either way, thanks for the fix!

PR #3000, opened 2026-07-24
Manually pack 24-bit RGB in GetPixel

khang06:
This 1-byte out-of-bounds read was causing random crashes when copying images to the clipboard with the WinForms backend, but probably happens everywhere else. Also, it probably led to problems for ARM due to unaligned reads, but I didn't check that.


khang06:
No idea if that was a repo setting since I have those automatic reviews disabled...
image

cwensley:
Thanks for the fix!  I've added some unit tests to verify and prevent regressions.

PR #3003, opened 2026-07-25
Mac: Fix ToEtoKeyEventArgs for alternative keyboard layouts

### LINKED ISSUE #3002: Mac: ToEtoKeyEventArgs doesn't respect non-QWERTY keyboard layout
This line is the issue:

https://github.com/picoe/Eto/blob/27316b9ace278140cb876c0a265202c614280389/src/Eto.Mac/MacConversions.cs#L414

NSEvent.KeyCode represents the physical key, it doesn't take into account the keyboard layout. Better would be roughly this:

```c#
Keys key = KeyMap.Convert(theEvent.CharactersIgnoringModifiers, 0);
if (key == Keys.None) {
  // Non-character keys like F1
  key = ... old logic;
}
```

cyanfish:
Fixes #3002

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,10 +3,12 @@
 Eto.Forms cross-platform UI toolkit. Keep this file compact — it loads every
 session. Record only durable, non-obvious Eto facts; skip anything obvious from code/git.
 
-**Standing instruction:** whenever you discover a non-obvious fact about this repo — a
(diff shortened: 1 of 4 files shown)


cwensley:
Thanks for the fix! This was definitely a problem with the Mac implementation of key events.  I've added some tests and fixes for various edge cases, and a few keys that didn't have inverse mappings.

PR #3005, opened 2026-07-27
Don't use TryGetData<object>

### LINKED ISSUE #3004: WinForms: DataObject regression in 2.11.0
`Clipboard.Instance.GetData` fails due to the following exception:

> System.NotSupportedException: 'System.Object' is not a concrete type, and could result in an unbounded deserialization attempt. Please use a concrete type or alternatively define a 'resolver' function that supports types that you are expecting to retrieve from the clipboard or use in drag-and-drop operations.

This is swallowed [here](https://github.com/picoe/Eto/blob/27316b9ace278140cb876c0a265202c614280389/src/Eto.Wpf/Forms/DataObjectHandler.cs#L311) so it's not actually thrown (found via debugging), but it falls back to manual stream reading which doesn't work for everything.

Regression introduced in https://github.com/picoe/Eto/commit/d0f1f40ae88077a50703234d2b4bc4b972c7cdde (net10-only use of `swf.Clipboard.TryGetData<object>(type, out var data);`).

cyanfish:
This fails as it requires a concrete type. For now keep using the obsolete method and suppress the warnings.

Fixes #3004

### DIFF
--- a/src/Eto.WinForms/Forms/ClipboardHandler.cs
+++ b/src/Eto.WinForms/Forms/ClipboardHandler.cs
@@ -105,12 +105,9 @@ namespace Eto.WinForms.Forms
 
 		protected override object InnerGetData(string type)
 		{
-#if NET10_0_OR_GREATER
-			swf.Clipboard.TryGetData<object>(type, out var data);
-			return data;
-#else
+#pragma warning disable WFDEV005
 			return swf.Clipboard.GetData(type);
-#endif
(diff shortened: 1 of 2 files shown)


cwensley:
Ah yes, that definitely won't work.. thanks for the fix.  It would be nice if `DataObject.GetObject<T>()` would use that API, but that can be done with a separate PR.

PR #3044, opened 2026-09-03
Fix .NET 10 Windows template target frameworks

### LINKED ISSUE #2931: WPF: `-windows` suffix is missing in `TargetFramework` in the generated `Wpf.csproj` file with `-f net10.0`
**Describe the bug**

When the target framework is net10.0, in the generated WPF project file, `-windows` is missing in `TargetFramework`. There is no issue when `-f net8.0` is given. This appears to originate from the [following lines in src/Eto.Forms.Templates/content/App-CSharp/.template.config/template.json](https://github.com/picoe/Eto/blob/e07b33dab6e4ad9e77985998a185c3fb27cc585b/src/Eto.Forms.Templates/content/App-CSharp/.template.config/template.json#L128-L131).

```json
		"IsNetCore": {
			"type": "computed",
			"value": "(Framework >= \"net5\")"
		},
```

It seems that "net10.0" < "net5" accidentally holds, while "net8.0" >= "net5". 

**To Reproduce**
1. Install .NET 10.0 SDK
2. Install Eto.Forms 2.11.0 template
3. Create a project without the explicit `-f` option (defaults to `net10.0`) or with `-f net10.0`.
4. `TargetFramework` in the generated `....Wpf.csproj` is set to `net10.0` instead of `net10.0-windows`. 

**Expected behavior**

`TargetFramework` in the generated `....Wpf.csproj` is set to `net10.0-windows`. 

**Actual behavior**

`TargetFramework` in the generated `....Wpf.csproj` is set to `net10.0`. 

```console
$ mkdir Sample
$ cd Sample
$ dotnet new etoapp -f net10.0
...
$ grep TargetFramework Sample.Wpf/Sample.Wpf.csproj
    <TargetFramework>net10.0</TargetFramework>
```

**Specifications**

  - Eto Version: 2.11.0
  - Platform(s): WPF
  - Operating System(s): macOS 26.4.1/Windows 11

pttydou:
Fixes #2931.

The template's `IsNetCore` expression compared framework names lexicographically, so `net10.0` was classified as a .NET Framework target. This change treats the three supported .NET Framework TFMs as the legacy cases and applies the same fix to the C#, F#, and VB templates.

Tests:
- Generated all eight supported frameworks for C#, F#, and VB from the source templates and checked the WPF and WinForms TFMs (48 assertions).
- Built `Eto.Forms.Templates` in Release (0 warnings, 0 errors).
- Installed the generated nupkg and smoke-tested net10.0 and net48 for all three languages.

### DIFF
--- a/src/Eto.Forms.Templates/content/App-CSharp/.template.config/template.json
+++ b/src/Eto.Forms.Templates/content/App-CSharp/.template.config/template.json
@@ -127,7 +127,7 @@
 		},
 		"IsNetCore": {
 			"type": "computed",
-			"value": "(Framework >= \"net5\")"
+			"value": "(Framework != \"net48\" && Framework != \"net472\" && Framework != \"net462\")"
 		},
 		"IsWindow": {
 			"type": "computed",
--- a/src/Eto.Forms.Templates/content/App-FSharp/.template.config/template.json
+++ b/src/Eto.Forms.Templates/content/App-FSharp/.template.config/template.json
@@ -127,7 +127,7 @@
 		},
 		"IsNetCore": {
 			"type": "computed",
-			"value": "(Framework >= \"net5\")"
+			"value": "(Framework != \"net48\" && Framework != \"net472\" && Framework != \"net462\")"
 		},
 		"IsWindow": {
 			"type": "computed",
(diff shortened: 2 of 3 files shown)


cwensley:
Hey @pttydou, thanks for this!  However, I went a different route for this in #3051 as it needs to check the TargetFrameworkOverride too, which is what is used when generating templates from VS.

